PR #1, opened 2025-01-20
2: Cleanup temp files

josephdrichard:
Config and socket files can all be deleted when applying a new ptp profile, and will be recreated by the new profile.


aneeshkp (approved):
/lgtm

PR #5, opened 2025-01-30
Fix DPLL Pin API

vitus133:
https://github.com/k8snetworkplumbingwg/ptp-operator/issues/5
Fix entity relations in the PIN API.
Add pin setting API (as a preparation phase for T-BC implementation)
/cc @aneeshkp  @josephdrichard


josephdrichard:
Can you rebase your branch onto main?

PR #9, opened 2025-02-06
Fix DPLL pin API

vitus133:
https://github.com/k8snetworkplumbingwg/ptp-operator/issues/5
Fix entity relations in the PIN API.
Add pin setting API (as a preparation phase for T-BC implementation)
This is instead of the #5 (now closed)

/cc @josephdrichard @aneeshkp


aneeshkp:
@vitus133  please rebase

josephdrichard (approved):

PR #10, opened 2025-02-06
keep in holdover until locked-holdover-acquired

vitus133:
https://github.com/k8snetworkplumbingwg/ptp-operator/issues/9

Simplify DPLL eventing logic by decoupling holdover and LHAQ states.
Remove "locked" DPLL state handling, keeping in Holdover until the LHAQ state reached.
Fix a corner case when holdover timer keeps running when DPLL goes to unlocked
Fix a race condition in tests where event source was not fast enough to be set before the event processing 

/cc @aneeshkp @josephdrichard

### DIFF
--- a/pkg/dpll/dpll.go
+++ b/pkg/dpll/dpll.go
@@ -359,7 +359,7 @@ func (d *DpllConfig) nlUpdateState(devices []*nl.DoDeviceGetReply, pins []*nl.Pi
 				glog.Info("discarding on invalid lock status: ", nl.GetDpllStatusHR(reply))
 				continue
 			}
-			glog.Info(nl.GetDpllStatusHR(reply))
+			glog.Info(nl.GetDpllStatusHR(reply), " ", d.iface)
 			switch nl.GetDpllType(reply.Type) {
 			case "eec":
 				d.frequencyStatus = int64(reply.LockStatus)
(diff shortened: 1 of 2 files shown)


pkg/dpll/dpll.go
```diff
@@ -627,47 +628,62 @@ func (d *DpllConfig) MonitorDpll() {
...
 	dpllStatus := d.getWorseState(d.phaseStatus, d.frequencyStatus)
+
 	switch dpllStatus {
 	case DPLL_FREERUN, DPLL_INVALID, DPLL_UNKNOWN:
 		d.inSpec = false
 		d.sourceLost = true
-		if d.hasGNSSAsSource() && d.onHoldover {
-			d.holdoverCloseCh <- true
+		glog.Infof("on holdover %t has GNSS source %t", d.onHoldover, d.hasGNSSAsSource())
+		if d.onHoldover {
```
aneeshkp, line 637: we need to gate this for d.hasGNSSAsSource() ; if  this clock id has GNSS only then we control holdover

aneeshkp: we don't close holdover once its in holdover if other nic state change to Freerun etc

pkg/dpll/dpll.go
```diff
@@ -677,17 +693,10 @@ func (d *DpllConfig) stateDecision() {
...
-				d.state = event.PTP_HOLDOVER
-				go d.holdover()
-			}
-			return // do not send event holdover  will handle it
-		case !d.inSpec: // this is for GNSS only
-			glog.Infof("dpll is not in spec, state is DPLL_LOCKED_HO_ACQ or DPLL_HOLDOVER, offset is out of range, state is FREERUN(%s)", d.iface)
+		} else {
+			glog.Infof("dpll is not in spec, state is DPLL_LOCKED_HO_ACQ, offset is out of range, state is FREERUN(%s)", d.iface)
 			d.state = event.PTP_FREERUN
+			d.inSpec = true
```
aneeshkp, line 699: if state is freeRUN , how is  inSpec is true

pkg/dpll/dpll.go
vitus133: Fixed, thanks!

pkg/dpll/dpll.go
vitus133: Indeed, fixed

aneeshkp (approved):
/lgtm

PR #15, opened 2025-02-12
Add readiness endpoint

nocturnalastro:


pkg/daemon/daemon.go
```diff
@@ -1009,6 +1029,7 @@ func (p *ptpProcess) processPTPMetrics(output string) {
 		p.ProcessSynceEvents(logEntry)
 	} else {
 		configName, source, ptpOffset, clockState, iface := extractMetrics(p.messageTag, p.name, p.ifaces, output)
+		p.hasCollectedMetrics = true
```
aneeshkp, line 1032: you would know if there are metrics only if you have ptpOffset, I think That coould be good start. But if events are enable dthen extract metrics is moved to cloud-event-proxy and it will not enter into this condition .

aneeshkp (dismissed):

aneeshkp (approved):
/lgtm

PR #19, opened 2025-03-21
Wait for socket to be ready before starting the command

edcdavid:
When events are enabled, logs from ptp4l and other commands are sent to the cloud-event-proxy container via a socket. This PR delays the start of the command until after the socket is setup.
Otherwise, the early logs are missed, so early state transitions used in metrics are lost


aneeshkp (dismissed):
/lgtm

aneeshkp (approved):
/lgtm

PR #25, opened 2025-04-04
Use clockType instead of profileClockType

nocturnalastro:
clockType will default from inspecting the
ptp4l config

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -183,7 +183,6 @@ type ptpProcess struct {
 	c                   *net.Conn
 	hasCollectedMetrics bool
 	trIfaceName         string // Time receiver interface name for T-BC clock monitoring
-	profileClockType    string
 }
 
 func (p *ptpProcess) Stopped() bool {
@@ -424,6 +423,7 @@ func (dn *Daemon) applyNodePTPProfiles() error {
 				for _, d := range p.depProcess {
 					if d != nil {
 						time.Sleep(3 * time.Second)
+						glog.Infof("Starting %s", d.Name())
 						go d.CmdRun(false)
 						time.Sleep(3 * time.Second)
 						dn.pluginManager.AfterRunPTPCommand(&p.nodeProfile, d.Name())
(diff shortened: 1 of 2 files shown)


pkg/daemon/daemon.go
```diff
@@ -598,9 +598,9 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 		if found {
 			switch profileClockType {
 			case "T-GM":
-				clockType = "GM"
+				clockType = event.GM
```
aneeshkp, line 601: this should default to T-GM, older T-GM profiles don't have  this value

nocturnalastro: It does on 603 that was the issue.

aneeshkp (dismissed):
/lgtm

pkg/daemon/daemon.go
```diff
@@ -513,7 +513,37 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
...
+		case "T-BC":
+			clockType = event.BC
+		default:
+			clockType = event.ClockUnset
+		}
+	} else {
+		clockType = event.ClockUnset
+	}
+
+	origionalOutput := &ptp4lConf{}
```
josephdrichard, line 531: spelling

pkg/daemon/daemon.go
```diff
@@ -513,7 +513,37 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
...
+	err = origionalOutput.populatePtp4lConf(configInput)
```
josephdrichard, line 532: How is configInput not empty here?

pkg/daemon/daemon.go
aneeshkp: how come linter didn't catch this ?

pkg/daemon/daemon.go
```diff
@@ -675,16 +683,16 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 			ptpClockThreshold: getPTPThreshold(nodeProfile),
 			haProfile:         haProfile,
 			syncERelations:    relations,
-			profileClockType:  profileClockType,
 		}
+
 		if pProcess == ptp4lProcessName {
-			if profileClockType == "T-BC" {
+			if clockType == event.BC {
 				dprocess.trIfaceName, _ = (*nodeProfile).PtpSettings["upstreamPort"]
```
aneeshkp, line 690: This will break , since this is not applied for regular BC and should use check for  key/value availability  before setting

pkg/daemon/daemon.go
nocturnalastro: Yeah, thanks for spotting this. Also looks like populatePtp4l is parsing more than ptp4l. I'll general parsing back to the loop. And I'll call it on `nodeProfile.Ptp4lConf` here.

pkg/daemon/daemon.go
```diff
@@ -675,16 +690,16 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 			ptpClockThreshold: getPTPThreshold(nodeProfile),
 			haProfile:         haProfile,
 			syncERelations:    relations,
-			profileClockType:  profileClockType,
 		}
+
 		if pProcess == ptp4lProcessName {
-			if profileClockType == "T-BC" {
-				dprocess.trIfaceName, _ = (*nodeProfile).PtpSettings["upstreamPort"]
+			if port, found := (*nodeProfile).PtpSettings["upstreamPort"]; found && clockType == event.BC {
```
aneeshkp, line 696: found shadows declaration at line 517

josephdrichard (approved):

PR #26, opened 2025-04-04
Fix clock type

edcdavid:
Not sure about the full scope of the clock type, but eems to be used only in T-GM, only if it is not defined in the profile. If it is not present in the profile, the clockType is inferred from all configurations for all processes(phc2sys, ts2phc, ptp4l, etc). However, only the ptp4l configuration can actually be used to infer the clock type.
Modified the following:
- Infer clock type from ptp4l only, ptp4l is processed first so other processe can get the clock type
- After discussing with @aneeshkp and @nocturnalastro, I changed the logic to identify OC clocks properly (including dual follower).

### DIFF
--- a/pkg/daemon/config.go
+++ b/pkg/daemon/config.go
@@ -37,7 +37,7 @@ type ptp4lConfSection struct {
 type ptp4lConf struct {
 	sections         []ptp4lConfSection
 	mapping          []string
-	profile_name     string
+	profileName      string
 	clock_type       event.ClockType
 	gnss_serial_port string // gnss serial port
 }
@@ -119,7 +119,9 @@ func (output *ptp4lConf) populatePtp4lConf(config *string) error {
 	var currentSection ptp4lConfSection
 	output.sections = make([]ptp4lConfSection, 0)
 	globalIsDefined := false
-	hasSlaveConfigDefined := false
+	isSlaveOnly := false
+	atLeastOneMaster := false
+	allMasters := true
 
 	if config != nil {
 		for _, line := range strings.Split(*config, "\n") {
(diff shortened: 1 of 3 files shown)


pkg/daemon/config.go
```diff
@@ -142,14 +144,17 @@ func (output *ptp4lConf) populatePtp4lConf(config *string) error {
...
-					currentSection.options[line[:split]] = line[split:]
-					if (line[:split] == "masterOnly" && line[split:] == "0") ||
-						(line[:split] == "serverOnly" && line[split:] == "0") ||
-						(line[:split] == "slaveOnly" && line[split:] == "1") ||
-						(line[:split] == "clientOnly" && line[split:] == "1") {
-						hasSlaveConfigDefined = true
+				split := strings.Split(line, " ")
+				if len(split) == 2 {
+					currentSection.options[split[0]] = split[1]
+					if currentSection.options["slaveOnly"] == "1" && globalIsDefined {
```
nocturnalastro, line 150: `slaveOnly` is deprecated according to this [doc](https://linuxptp.nwtime.org/documentation/ptp4l/) we should also check `clientOnly`

pkg/daemon/config.go
```diff
@@ -142,14 +144,17 @@ func (output *ptp4lConf) populatePtp4lConf(config *string) error {
...
+						isSlaveOnly = true
+					}
+					if currentSection.options["masterOnly"] == "1" {
```
nocturnalastro, line 153: `masterOnly` is deprecated according to this [doc](https://linuxptp.nwtime.org/documentation/ptp4l/) we should also check `serverOnly`

nocturnalastro (changes requested):

PR #33, opened 2025-04-30
Fix metrics alias

nocturnalastro:


pkg/daemon/metrics.go
```diff
@@ -783,8 +783,12 @@ func (m *masterOffsetInterface) set(configName string, value string) {
 	defer m.Unlock()
 	alias := ""
 	if value != "" {
-		r := []rune(value)
-		alias = string(r[:len(r)-1]) + "x"
+		dotIndex := strings.Index(value, ".")
```
aneeshkp, line 786: can you add examples  in comment section around this pls

nocturnalastro: I've interlaced the comments instead of putting them in a block at the top let me know if you'd prefer the other way. 

Once you're happy I'll make the same change in cloud event proxy :)

aneeshkp: ensfx.100.20 etc

aneeshkp (approved):
/lgtm

rdavid (dismissed):

aneeshkp (approved):
/lgtm

PR #34, opened 2025-05-02
Variable rate log reduction

josephdrichard:


aneeshkp (commented):
The refactor is significant ,
Please consider splitting some parts ( filterOutput, parseOffsetStats) into a separate file or package , the  daemon grows further.

The logFilter struct and its logic (especially filterOutput) could use more inline comments to help future maintainers understand the flow.

pkg/logfilter/logfilter.go
```diff
@@ -0,0 +1,197 @@
...
+
+	filter.expiryTime = time.Now().Add(filter.batchLength)
+	filter.summaryText = summaryText
+	filter.summaryFields = summaryFields
+	for _, reducer := range reducers {
+		reducerRegex, reducerRegexErr := regexp.Compile(reducer)
+		if reducerRegexErr != nil {
+			glog.Infof("Failed parsing reducer %s: %d.", reducer, reducerRegexErr)
+		}
+		filter.logFilterReducerRegexes = append(filter.logFilterReducerRegexes, reducerRegex)
```
aneeshkp, line 65: you should not appedn if reducerRegexErr is not niil

PR #37, opened 2025-05-21
Adding gated ptp-operator pre-merge CI

edcdavid:
This PR runs the ptp CI tests added to ptp-operator but runs them before merging.
- To start the job, an ok-to-test label is required. Only members with write permission on the repository are allowed to set labels
- a member with write access sets the ok-to-test label
- the ok-to-test label is immediately removed before the tests are started in aws
- the tests are started in aws and have 30 minutes to complete.
If the used wants to run the tests again when pushing changes, he would need again someone with write access to set the label
Members with write access can just set the label themselves to retest.

### DIFF
--- a/.github/workflows/aws-ci.yaml
+++ b/.github/workflows/aws-ci.yaml
@@ -1,0 +1,107 @@
+name: CI tests AWS
+on:
+  pull_request_target:
+    types: [labeled, synchronize]  # Triggered when the PR is labeled or synchronized
+permissions:
+  id-token: write
+  contents: read
+  pull-requests: write
+jobs:
+  deploy-ec2:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Remove ok-to-test label if new commit
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const label = "ok-to-test";
+            const pr = context.payload.pull_request;
+            const labels = pr.labels.map(l => l.name);
+            if (labels.includes(label)) {
+              await github.rest.issues.removeLabel({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: pr.number,
+                name: label
+              });
+              console.log(`Removed ${label} due to new push`);
+            }
+      - name: Checkout the code
+        uses: actions/checkout@v3
+        with:
+          repository: k8snetworkplumbingwg/ptp-operator
+          ref: main
+          path: ptp-operator
+      
+      - name: checkout the Current SHA for the linuxptp-daemon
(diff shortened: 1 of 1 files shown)


.github/workflows/aws-ci.yaml
```diff
@@ -0,0 +1,107 @@
...
+    timeout-minutes: 30
+    steps:
+      - name: Remove ok-to-test label if new commit
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const label = "ok-to-test";
+            const pr = context.payload.pull_request;
+            const labels = pr.labels.map(l => l.name);
+            if (labels.includes(label)) {
```
aneeshkp, line 21: this will remove ok-to-test for any label added by user?

aneeshkp: Add logic to remove ok-to-test only when the event is synchronize (i.e., new commit push), not when any label is added.

.github/workflows/aws-ci.yaml
```diff
@@ -0,0 +1,107 @@
+name: CI tests AWS
```
aneeshkp, line 1: I don't see where you check for trusted or member is the author of the label ok-to-test >

.github/workflows/aws-ci.yaml
```diff
@@ -0,0 +1,107 @@
...
+            --image-id ami-0f7270e9a0e492eac \
+            --subnet-id subnet-04ee218fe8f32cad5 \
+            --instance-type m6a.xlarge \
+            --security-group-ids sg-05e700edc2ed0a2d5 \
+            --tag-specifications 'ResourceType=instance,Tags=[{Key=Name,Value=ptp-operator PR-${{ github.event.number }}}]' \
+            --query 'Instances[0].InstanceId' \
+            --key-name ptp-sshkey-1 \
+            --output text)
+
+          echo "instance_id=$INSTANCE_ID" >> $GITHUB_ENV
```
aneeshkp, line 64: you are saving instance_id in to github env and then you never use it 
why don't you use ${{ env.instance_id }} instead?
Don't you need to re export?
instance_id=${{ env.instance_id }}

edcdavid: I do use instance_id later on when deleting the instance see:
```
      - name: Cleanup - Delete EC2 instance
        if: always()  # Ensure cleanup runs even if previous steps fail
        run: |
            echo "Cleaning up EC2 INSTANCE_ID=$instance_id"
            aws ec2 terminate-instances --instance-ids $instance_id
```

aneeshkp (approved):

PR #40, opened 2025-05-22
LinuxPTP Feature Flag Management

nocturnalastro:
LinuxPTP Feature Flag Management:
This module introduces a version-aware feature flag system to dynamically enable or disable capabilities in the linuxptp-daemon based on the installed linuxptp package version. It ensures consistent behavior across releases while simplifying support for new and deprecated features.

Dynamically detect the linuxptp version at runtime.

Enable a predictable set of feature flags based on version thresholds.

Provide a clean API (IsFeatureAvailable, AreFeaturesAvailable) for consumers.

Improve testability and maintainability by decoupling feature logic from business logic.

Allow safe deprecation or rollback of feature usage based on version changes.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/config"
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/daemon"
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/features"
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/leap"
 	ptpv1 "github.com/k8snetworkplumbingwg/ptp-operator/api/v1"
 	ptpclient "github.com/k8snetworkplumbingwg/ptp-operator/pkg/client/clientset/versioned"
@@ -125,6 +126,18 @@ func main() {
 	defer close(lm.Close)
 
 	tracker := &daemon.ReadyTracker{}
+
+	version := features.GetLinuxPTPPackageVersion()
+	ocpVersion := os.Getenv("BUILD_VERSION")
+	if ocpVersion == "" {
+		ocpVersion = features.LatestOCPInMatrix
+	} else {
+		ocpVersion, _ = strings.CutPrefix(ocpVersion, "v")
+		ocpVersion = strings.Join(strings.Split(ocpVersion, ".")[:2], ".")
+	}
+	// TODO: version needs to be sent to cloud event proxy when we intergrate feature flags there
+	features.SetFlags(version, ocpVersion)
+	features.Flags.Print()
 
 	go daemon.New(
 		nodeName,
--- a/pkg/features/README.md
+++ b/pkg/features/README.md
@@ -1,0 +1,22 @@
+# Feature Flags
+
+This package provides a central point of truth for the codebase, indicating which features are available for execution.
+It determines feature availability based on the installed version of the linuxptp package.
+Integrating this package allows for conditional code execution depending on the enabled features.
+
+## Example
+
+```go
+import (
+    "github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/features"
+)
+
+func Foo() {
+    if features.Flags.GM.Enabled {
+        fmt.Println("Generic GM stuff here")
+        if features.Flags.GM.Holdover {
+            fmt.Println("Holdover GM stuff here")
+        }
+    }
+}
+```
--- a/pkg/features/featureflags.go
+++ b/pkg/features/featureflags.go
@@ -1,0 +1,117 @@
+package features
+
+import (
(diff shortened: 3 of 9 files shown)


pkg/features/featureflags.go
```diff
@@ -0,0 +1,180 @@
...
+	dualBC      bool
+	twoPortOC   bool
+	ha          bool
+	leapFile    bool
+	bcHoldover  bool
+}
+
+// Versions ...
+const (
+	LinuxPTPVersion3112 = "3.1.1-2.el8_6.3"
```
aneeshkp, line 36: there is a possible this will bump minor version

nocturnalastro: That's why I added the comparison lib stuff, as we would not have to track each version. 

However we do have to maintain the comparison lib so that's the trade off.

pkg/features/featureflags.go
```diff
@@ -0,0 +1,165 @@
...
+var getVersion func() string
+
+func init() {
+	Flags = &Features{}
+	getVersion = getLinuxPTPPackageVersion
+}
+
+// Features ...
+type Features struct {
+	oc          bool
```
jzding, line 21: it would be nice to have a short description for each feature. For example the feature title from Jira (not include Jira number).

pkg/features/featureflags.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Check if version >= 4.4-1.el9
+	if version.Compare(VersionLinuxPTP441) >= 0 {
+		f.twoPortOC = true
+		f.bcHoldover = true
+	}
+}
+
+// IsDualBCAvaiable ...
+func (f *Features) IsDualBCAvaiable() bool {
```
josephdrichard, line 107: spelling

pkg/features/featureflags.go
nocturnalastro: Would be useful to get people who are more familiar with the features to write the descriptions.

cmd/main.go
```diff
@@ -126,6 +127,22 @@ func main() {
 
 	tracker := &daemon.ReadyTracker{}
 
+	version := features.GetLinuxPTPPackageVersion()
+	features.Flags.Init(version)
+	if stdoutToSocket {
+		// Write linuxptp version for cloud event proxy
+		err = os.WriteFile(
+			filepath.Join(cp.profileDir, "linuxptp-version.txt"),
```
edcdavid, line 135: minor: use const

pkg/features/featureflags.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Print ...
+func (f OCFeatures) Print() {
+	glog.Info("OC Enabled: ", f.Enabled)
+	glog.Info("OC DualPort: ", f.DualPort)
+}
+
+// And applies a logical and on the feature sets
+func (f OCFeatures) And(other OCFeatures) OCFeatures {
+	res := OCFeatures{}
```
aneeshkp, line 55: small change using direct struct for easy readbility , probbaly do it for all .  Will make it easier to read 
return OCFeatures{
		Enabled:  f.Enabled && other.Enabled,
		DualPort: f.DualPort && other.DualPort,
	}

pkg/features/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) error {
+    ...
+    if section.sectionName == "[global]" {
+        section.options["message_tag"] = messageTag
+        if socketPath != "" {
+            section.options["uds_address"] = socketPath
+        }
+
+        if feature.Flags.IsGMAvailable() { // Only runs if Grand master is enabled
```
aneeshkp, line 31: probably not a good example, configs parsing is done by looking what is given in ptpconfig
just have empty method
if feature.Flags.IsGMAvailable() { // Only runs if Grand master is enabled

pkg/features/ocp.go
```diff
@@ -0,0 +1,139 @@
+package features
+
+var ocpFeatureMatrix = map[string]Features{
```
aneeshkp, line 3: could do this approach , use yaml 
4.12:
  OC:
    Enabled: true
  BC:
    Enabled: true

4.13:
  inherit: 4.12
  BC:
    DualPort: true

4.14:
  inherit: 4.13
  GM:
    Enabled: true

4.15:
  inherit: 4.14

4.16:
  inherit: 4.15
  BC:
    PTPHA: true
  GM:
    HoldOver: true
    SyncE: true
  LogSeverity: true

nocturnalastro: Do you mean using yaml anchor? 

Or adding a value to features to say grab the previous version, and then combine them at run time?

nocturnalastro: We decided that doing inheritance while preserving the ability to disable features added undesired complexity to the checking code.

aneeshkp (approved):

PR #41, opened 2025-05-26
Allow duplicate options in ptpconf

josephdrichard:


pkg/daemon/config.go
```diff
@@ -19,6 +19,13 @@ import (
 	ptpv1 "github.com/k8snetworkplumbingwg/ptp-operator/api/v1"
 )
 
+// predefined config section names
+const (
```
aneeshkp, line 23: @josephdrichard  is it possible to add unit test for this?

pkg/daemon/config.go
```diff
@@ -29,19 +36,50 @@ type LinuxPTPConfUpdate struct {
...
+	if !ok {
+		conf.sectionOptions[sectionName] = make([]ptp4lConfOption, 0)
+	}
+	if key == "" {
+		return
+	}
+	if overwrite {
+		for i := range conf.sectionOptions[sectionName] {
+			if conf.sectionOptions[sectionName][i].key == key {
+				conf.sectionOptions[sectionName][i] = ptp4lConfOption{key: key, value: value}
```
aneeshkp, line 75: return // missing!

aneeshkp (commented):
When you are here , can you change the comparison from string to byte
func (l *LinuxPTPConfUpdate) UpdateConfig(nodeProfilesJson []byte) error {
	if string(l.appliedNodeProfileJson) == string(nodeProfilesJson) {
		return nil
	}
	
	Instead use this , I had updated  this in the sidecar
	
	if bytes.Equal(l.appliedNodeProfileJSON, nodeProfilesJSON) {
		return true, nil
	}

pkg/daemon/config.go
```diff
@@ -19,6 +19,13 @@ import (
 	ptpv1 "github.com/k8snetworkplumbingwg/ptp-operator/api/v1"
 )
 
+// predefined config section names
+const (
+	GlobalSectionName  = "[global]"
+	NmeaSectionName    = "[nmea]"
+	UnicastSectionName = "[unicast_master_table]"
+)
```
aneeshkp, line 27: no considering other sections, for an example in syncE ?
[global]
logging_level               7
use_syslog                  0
verbose                     1
message_tag [synce4l.0.config]
smc_socket_path             /run/synce4l_socket
[<synce1>]
module_name                 ice
dnu_prio  0XFF
network_option              2
extended_tlv                1
recover_time                60
clock_id  5799633565433967748
[ens7f0]
tx_heartbeat_msec           1000
rx_heartbeat_msec           10
allowed_qls                 0xE
allowed_ext_qls             0x40
[{GNSS}]
board_label                 GNSS-1PPS
input_QL                    0x1
input_ext_QL                0x20

aneeshkp: `[global]
logging_level               7
use_syslog                  0
verbose                     1
message_tag [synce4l.0.config]
smc_socket_path             /run/synce4l_socket
[<synce1>]
module_name                 ice
dnu_prio  0XFF
network_option              2
extended_tlv                1
recover_time                60
clock_id  5799633565433967748
[ens7f0]
tx_heartbeat_msec           1000
rx_heartbeat_msec           10
allowed_qls                 0xE
allowed_ext_qls             0x40
[{GNSS}]
board_label                 GNSS-1PPS
input_QL                    0x1
input_ext_QL                0x20
`

aneeshkp:
please add description and link to issue.

pkg/daemon/config.go
nocturnalastro: should this break once its found?

pkg/daemon/config.go
```diff
@@ -255,52 +315,71 @@ func (conf *ptp4lConf) extractSynceRelations() *synce.Relations {
...
 	var nmea_source event.EventSource
 
-	for _, section := range conf.sections {
-		configOut = fmt.Sprintf("%s\n%s", configOut, section.sectionName)
+	// Maintain consistent order of sections
+	sectionNames := make([]string, 0)
+	for sectionName := range conf.sectionOptions {
+		sectionNames = append(sectionNames, sectionName)
+	}
+	sort.Strings(sectionNames)
```
aneeshkp, line 360: why this sorting? will this sorting change  order of interface provided ?

pkg/daemon/daemon_test.go
```diff
@@ -470,3 +470,54 @@ func TestDaemon_ProcessSynceLogs(t *testing.T) {
...
+			testName:       "Basic OC test",
+			iface:          "ens2f0",
+			messageTag:     "test message_tag",
+			socketPath:     "/var/run/ptp4l.0.socket",
+			pProcess:       "",
+		},
+		{
+			profileName:    "BC",
+			testPtp4lConf:  "[global]\nslaveOnly 0\n[ens1f0]\nmasterOnly 0\n[ens1f1]\nmasterOnly 1",
+			expectedOutput: "#profile: BC\n\n[ens1f0]\nmasterOnly 0\n[ens1f1]\nmasterOnly 1\n[global]\nslaveOnly 0\nmessage_tag test message_tag\nuds_address /var/run/ptp4l.0.socket",
```
nocturnalastro, line 500: Can we do something like this I think for tests its better if its easier to read. 
```suggestion
			expectedOutput: `
	#profile: BC
	
	[ens1f0]
	masterOnly 0
	[ens1f1]
	masterOnly 1
	[global]
	slaveOnly 0
	message_tag test message_tag
	uds_address /var/run/ptp4l.0.socket`,
```

josephdrichard: > Can we do something like this I think for tests its better if its easier to read.

That was failing because of mismatched line endings.

pkg/daemon/config.go
josephdrichard: updated

pkg/daemon/daemon_test.go
nocturnalastro: Probably a leading newline, could use `strings.TrimSpace` on the the expected. As will remove the spaces from the start and the end where you wouldn't expect the rendered output to have any but I'm not that bothered just thought it might make the test cases easer to understand.

aneeshkp (approved):

PR #46, opened 2025-06-10
Build log parsing package

aneeshkp:
# ptp log parser 
### 1. Metrics Extraction
- Time offset measurement
- Frequency adjustment tracking
- Path delay calculation
- Maximum offset monitoring
- Clock state tracking

### 2. State Management
- Interface role tracking
- Port state monitoring
- Fault detection
- Configuration state maintenance

### 3. Event Processing
- Port state transition tracking
- Role change detection
- Fault state monitoring
- Synchronization status updates

### 4. Error Handling
- Invalid log line detection
- Missing field handling
- Invalid numeric value processing
- State transition validation
- Port number validation

## Use Cases

1. **PTP Synchronization Monitoring**
   - Track time offset between master and slave
   - Monitor frequency adjustment
   - Measure path delay
   - Detect synchronization issues

2. **Port State Management**
   - Track port role changes
   - Monitor state transitions
   - Detect faulty states
   - Manage interface configuration

3. **System Clock Synchronization**
   - Monitor PHC to system clock synchronization
   - Track CLOCK_REALTIME adjustments
   - Measure system clock offset
   - Detect synchronization issues

4. **Metrics Collection**
   - Extract timing metrics
   - Track state changes
   - Monitor synchronization status
   - Collect performance data

## Integration
The package is designed to be integrated with:
- Monitoring systems
- Metrics collection frameworks
- Log processing pipelines
- State management systems

## Dependencies
- Go standard library
- PTP daemons (ptp4l, phc2sys)
- Linux PTP stack

## Future Enhancements
1. Support for additional PTP daemons
2. Enhanced metrics collection
3. Improved state management
4. Extended event processing
5. Additional log format support

### DIFF
--- a/pkg/parser/base.go
+++ b/pkg/parser/base.go
@@ -1,0 +1,77 @@
+package parser
+
+import (
+	"fmt"
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/parser/state"
+	"strings"
+)
+
+// MetricsExtractor is an interface for extracting metrics from log lines.
+type MetricsExtractor interface {
+	ProcessName() string
+	Extract(logLine string) (*Metrics, error)
+}
+
+// BaseMetricsExtractor is a base struct for all metrics extractors.
+// It provides common functionality for extracting metrics from log lines.
+type BaseMetricsExtractor struct {
+	ProcessNameStr   string                                  // Process name (e.g., "ptp4l", "phc2sys")
+	ExtractSummaryFn func(logLine string) (*Metrics, error)  // Function to extract summary metrics
+	ExtractRegularFn func(logLine string) (*Metrics, error)  // Function to extract regular metrics
+	ExtraEventFn     func(logLine string) (*PTPEvent, error) // Function to extract events
+	State            *state.SharedState                      // Shared state for tracking PTP configuration
+}
+
+// ProcessName returns the name of the process that is being extracted.
+func (b *BaseMetricsExtractor) ProcessName() string {
+	return b.ProcessNameStr
+}
+
+// Extract extracts metrics from a log line.
+// It determines the type of log line and calls the appropriate extraction function.
+func (b *BaseMetricsExtractor) Extract(output string) (*Metrics, *PTPEvent, error) {
+	if output == "" {
+		return nil, nil, fmt.Errorf("empty log line")
+	}
+
+	output = removeMessageSuffix(output)
+
+	// Determine log line type and extract accordingly
+	if strings.Contains(output, " max ") {
+		metrics, err := b.ExtractSummaryFn(output)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to extract summary metrics: %v", err)
+		}
+		return metrics, nil, nil
+	} else if strings.Contains(output, " offset ") {
+		metrics, err := b.ExtractRegularFn(output)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to extract regular metrics: %v", err)
+		}
+		return metrics, nil, nil
+	} else if b.ExtraEventFn != nil {
+		event, err := b.ExtraEventFn(output)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to extract event: %v", err)
+		}
+		return nil, event, nil
+	}
+
+	return nil, nil, fmt.Errorf("log output is not parsable: %s", output)
+}
+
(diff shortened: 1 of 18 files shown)


pkg/parser/base.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+type MetricsExtractor interface {
+	ProcessName() string
+	Extract(configName, logLine string, ifaces config.IFaces) (error, *Metrics)
+}
+
+type BaseMetricsExtractor struct {
+	ProcessNameStr   string
+	ExtractSummaryFn func(messageTag, configName, logLine string, ifaces config.IFaces, _ shared.SharedState) (error, *Metrics)
```
nocturnalastro, line 17: Do interfaces, messageTag, configName change? Can they just be part of the struct on creation?

pkg/parser/base.go
```diff
@@ -0,0 +1,60 @@
...
+	ExtractRegularFn func(messageTag, configName, logLine string, ifaces config.IFaces, _ shared.SharedState) (error, *Metrics)
```
nocturnalastro, line 18: Do we need separate functions? for these or can we inspect the log line and dispatch there?

pkg/parser/phc2sys_parser.go
```diff
@@ -0,0 +1,206 @@
...
+		return nil, nil
+	}
+
+	maxPtpOffset, err = strconv.ParseFloat(fields[3], 64)
+	if err != nil {
+		err = fmt.Errorf("%s failed to parse max offset from the output %s error %v", processName, fields[1], err)
+		return nil, nil
+	}
+
+	switch fields[4] {
```
nocturnalastro, line 164: Should this be its own function for reuse as seems to be a common thing for linuxptp

pkg/parser/phc2sys_parser.go
```diff
@@ -0,0 +1,212 @@
...
+		Delay:      delay,
+		ClockState: "", // Summary metrics don't include clock state
+		Source:     master,
+	}
+}
+
+// extractRegularPhc2Sys parses regular metrics from phc2sys output
+// Expected format: ptp4l.0.config master offset -2162130 s2 freq +22451884 delay 374976
+func extractRegularPhc2Sys(messageTag, configName, output string, ifaces config.IFaces) (error, *Metrics) {
+	if output == "" {
```
edcdavid, line 134: Why not use a single regex to parse these strings in the metrics data structure, with no extra processing beyond augmenting the data. It would be nice also to reduce the number of log variation that we are trying to support. For example, why also support the summary format when it is missing the clockstate? Looking at the current code at https://github.com/aneeshkp/linuxptp-daemon-k8s/blob/86a286a948752259f996503f84bd25755dda6313/pkg/daemon/daemon.go#L1065 it seems that it would not behave properly unless non-summary is configured. Instead why not let the ptp-operator configure the proper logging, or report an warning/error?
For instance the non-summary phc2sys could be parsed with something like: 
```
var logRegex = regexp.MustCompile(
	`^(ptp4l\.\d+\.config)\s+master offset\s+(-?\d+)\s+(s\d)\s+freq\s+([+-]?\d+)\s+(?:path\s+)?delay\s+(\d+)`,
)

func parsePhc2sysLog(line string) (*LogPhc2sysEntry, error) {
	matches := logRegex.FindStringSubmatch(line)
	if len(matches) != 6 {
		return nil, fmt.Errorf("log does not match expected format: %q", line)
	}

	offset, err := strconv.Atoi(matches[2])
	if err != nil {
		return nil, fmt.Errorf("invalid master offset %q: %w", matches[2], err)
	}

	freq, err := strconv.Atoi(matches[4])
	if err != nil {
		return nil, fmt.Errorf("invalid frequency adjustment %q: %w", matches[4], err)
	}

	delay, err := strconv.Atoi(matches[5])
	if err != nil {
		return nil, fmt.Errorf("invalid delay %q: %w", matches[5], err)
	}

	var servo ServoState
	switch matches[3] {
	case "s0", "s1":
		servo = Freerun
	case "s2", "s3":
		servo = Locked
	default:
		return nil, fmt.Errorf("unknown servo state: %q", matches[3])
	}

	return &LogPhc2sysEntry{
		Interface:           configToInterface(matches[1]),
		MasterOffset:        offset,
		Servo:               servo,
		FrequencyAdjustment: freq,
		PathDelay:           delay,
	}, nil
}
```

PR #48, opened 2025-06-11
Increase timeout of upstream CI tests from 30mins to 60mins

edcdavid:
This is to support adding more tests, related to https://github.com/k8snetworkplumbingwg/ptp-operator/pull/61


aneeshkp (approved):
/lgtm

PR #53, opened 2025-06-18
Variable rate log reduction unit test

josephdrichard:


pkg/logfilter/logfilter_test.go
```diff
@@ -0,0 +1,111 @@
...
+type logFilterTestCase struct {
+	testName       string
+	processName    string
+	messageTag     string
+	ptpSettings    map[string]string
+	logs           []string
+	expectedOutput string
+}
+
+func initEnhancedLogFilterTestCases() []logFilterTestCase {
```
aneeshkp, line 21: Edge Cases: Consider tests for empty logs, logs with only one relevant line, or malformed log lines to further harden the logic.

aneeshkp: Also I would apperciate if you can add behaviour test for  expiryTime time and batchLength 
filter.expiryTime = time.Now().Add(filter.batchLength)

josephdrichard: Added test cases for malformed, empty, and only one line for both phc2sys and ptp4l filters.
I don't see a good way to test anything related to time, both because we always set it using time.Now(), which will give different values on different calls, and if we test anything that involves waiting, this will significantly increase how long our tests took to run.  I broke out actually doing the flushing the logs into a separate function, so that for testing it wouldn't be necessary to use time at all in our testing.

PR #56, opened 2025-06-20
Intial log parsing module

nocturnalastro:
This is an initial base for the log parsing lib. 

These are not yet integrated!
The plan is to iterate on these.


aneeshkp (approved):
/lgtm

PR #58, opened 2025-06-26
Fix gpspipe multiple process and named pipe creation issues

aneeshkp:
This PR improves process management and error handling for the gpsd and gpspipe daemons by:

Adding proper process termination with timeouts (graceful SIGTERM, then SIGKILL if needed)
Improving restart logic and reducing tight restart loops
Enhancing named pipe (FIFO) creation with retries and better error messages

Problem
The gpspipe process was experiencing two critical issues:
Multiple processes running simultaneously: The restart logic in CmdRun() was creating new processes before properly checking if the previous process should be stopped, leading to multiple gpspipe processes running at the same time.
Named pipe creation errors: The mkFifo() function was failing when a named pipe already existed, causing the error "Error creating named pipe, GNSS monitoring will not work as expected

Solution:
Added proper gp.Stopped() checks at the beginning of each loop iteration in CmdRun()
Enhanced mkFifo() function to check for existing named pipes and remove them before creation
Added graceful shutdown with SIGTERM signal.

Also implmented filteringStderrWriter to selectively suppressing known noisy errors like "Inappropriate ioctl for device" from being logged,  while still allowing all other stderr output to pass through.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -291,18 +291,7 @@ func (dn *Daemon) Run() {
 		case <-tickerPmc.C:
 			dn.HandlePmcTicker()
 		case <-dn.stopCh:
-			for _, p := range dn.processManager.process {
-				if p != nil {
-					for _, d := range p.depProcess {
-						if d != nil {
-							d.CmdStop()
-							d = nil
-						}
-					}
-					p.cmdStop()
-					p = nil
-				}
-			}
+			dn.stopAllProcesses()
 			glog.Infof("linuxPTP stop signal received, existing..")
 			return
 		}
@@ -358,29 +347,7 @@ func (dn *Daemon) applyNodePTPProfiles() error {
 	dn.readyTracker.setConfig(false)
 
 	glog.Infof("in applyNodePTPProfiles")
-	for _, p := range dn.processManager.process {
-		if p != nil {
-			glog.Infof("stopping process.... %s", p.name)
-			p.cmdStop()
-			if p.depProcess != nil {
-				for _, d := range p.depProcess {
-					if d != nil {
-						d.CmdStop()
-						d = nil
-					}
-				}
-			}
-			p.depProcess = nil
-			p.hasCollectedMetrics = false
-			//cleanup metrics
-			deleteMetrics(p.ifaces, p.haProfile, p.name, p.configName)
-			if p.name == syncEProcessName && p.syncERelations != nil {
-				deleteSyncEMetrics(p.name, p.configName, p.syncERelations)
-			}
-			p = nil
-		}
-	}
-
+	dn.stopAllProcesses()
 	// All process should have been stopped,
 	// clear process in process manager.
 	// Assigning processManager.process to nil releases
@@ -390,7 +357,7 @@ func (dn *Daemon) applyNodePTPProfiles() error {
 	dn.processManager.process = nil
 
 	// All configs will be rebuild, and sockets recreated, so they can all be deleted
-	dn.cleanupTempFiles()
+	_ = dn.cleanupTempFiles()
 
 	// TODO:
 	// compare nodeProfile with previous config,
@@ -680,10 +647,6 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) 
 				}
 				// TODO: move this to plugin or call it from hwplugin or leave it here and remove Hardcoded
 				gmInterface := dprocess.ifaces.GetGMInterface().Name
-
-				if e := mkFifo(); e != nil {
-					glog.Errorf("Error creating named pipe, GNSS monitoring will not work as expected %s", e.Error())
-				}
 
 				gpsDaemon := &GPSD{
 					name:        GPSD_PROCESSNAME,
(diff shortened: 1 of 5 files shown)


pkg/daemon/gpsd.go
```diff
@@ -116,28 +120,67 @@ func (g *GPSD) Stopped() bool {
...
+	g.setStopped(true)
+	g.ProcessStatus(nil, PtpProcessDown)
+
 	// Need to ensure cleanup to workaround ublox cpu spike bug
 	if g.ublxTool != nil {
 		g.ublxTool.UbloxPollStop()
 	}
+
 	if g.cmd == nil {
+		glog.Infof("No command to stop for %s", g.name)
```
josephdrichard, line 137: eh?

pkg/daemon/gpspipe.go
```diff
@@ -108,39 +132,205 @@ func (gp *gpspipe) CmdRun(stdoutToSocket bool) {
...
+	if !info.IsDir() {
+		return false
+	}
+
+	// Check if we have write permissions
+	if info.Mode().Perm()&0200 == 0 {
+		return false
+	}
+
+	// Try to create a temporary file to verify write access
```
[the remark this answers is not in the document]
aneeshkp: fixed it , not required to check permission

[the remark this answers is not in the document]
aneeshkp: test write is most reliable  check

pkg/daemon/gpsd.go
```diff
@@ -24,6 +25,17 @@ const (
 	GNSSMONITOR_INTERVAL = 1 * time.Second
 )
 
+type filteringStderrWriter struct{}
+
+func (w *filteringStderrWriter) Write(p []byte) (n int, err error) {
+	if bytes.Contains(p, []byte("Inappropriate ioctl for device")) {
+		// Suppress this error
+		return len(p), nil
```
nocturnalastro, line 33: Should this return 0? 
I'm not sure.

pkg/daemon/gpsd.go
```diff
@@ -176,18 +188,18 @@ func (g *GPSD) CmdRun(stdoutToSocket bool) {
 		g.ProcessStatus(nil, PtpProcessUp)
 		glog.Infof("Starting %s...", g.Name())
 		glog.Infof("%s cmd: %+v", g.Name(), g.cmd)
-		g.cmd.Stderr = os.Stderr
-		var err error
+		g.cmd.Stderr = &filteringStderrWriter{}
+		var cmdErr error
```
nocturnalastro, line 192: Not sure of the need for this change but its ok.

pkg/daemon/gpsd.go
aneeshkp: Great question . I am acting like AI :-P 
n = number of bytes successfully “written”
err is nill , might make cmd.StdoutPipe() think nothing was return nor error

pkg/daemon/gpsd.go
nocturnalastro: Oh I just an cmdErr vs err, I was being picky about the name change :P

pkg/daemon/gpsd.go
nocturnalastro: Ah ok that makes sense thanks :)

pkg/daemon/gpspipe.go
```diff
@@ -66,22 +69,43 @@ func (gp *gpspipe) Stopped() bool {
...
+		signalErr := gp.cmd.Process.Signal(syscall.SIGTERM)
+		if signalErr != nil {
 			glog.Errorf("Failed to send term signal to named pipe: %s", GPSPIPE_SERIALPORT)
 			return
 		}
+
+		// Wait for process to terminate with timeout
+		done := make(chan error, 1)
+		go func() {
+			done <- gp.cmd.Wait()
```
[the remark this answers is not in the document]
nocturnalastro: I did a go-routine and main in the go playground and he behaved fine.

pkg/daemon/gpsd.go
aneeshkp: fixed it

pkg/daemon/gpspipe.go
```diff
@@ -108,39 +132,205 @@ func (gp *gpspipe) CmdRun(stdoutToSocket bool) {
...
+			glog.Errorf("Failed to create named pipe after %d attempts. Last error: %v. GNSS monitoring will not work.", maxRetries, err)
+
+			// In test environment, return error instead of panicking
+			if os.Getenv("SKIP_GNSS_MONITORING") == "1" {
+				glog.Warning("In test environment, returning error instead of panicking")
+				return fmt.Errorf("failed to create named pipe after %d attempts: %v", maxRetries, err)
+			}
+
+			// In production, panic as this is critical for GNSS monitoring
+			glog.Fatalf("CRITICAL: Failed to create named pipe after %d attempts: %v. GNSS monitoring is essential and cannot continue.", maxRetries, err)
```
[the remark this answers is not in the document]
aneeshkp: this should never happen, let me see how often it fails at this point while we leave it in 4.20 an 4.19 branch and also I will ship dev build to customer

pkg/daemon/gpspipe.go
```diff
@@ -65,26 +69,62 @@ func (gp *gpspipe) Stopped() bool {
 
 // CmdStop ... stop gpspipe
 func (gp *gpspipe) CmdStop() {
+	defer func() {
+		// Clean up (delete) the named pipe, this ensures that the named pipe is deleted when the process is terminated
+		// no stale data is left in the named pipe
+		removeErr := os.Remove(GPSPIPE_SERIALPORT)
+		if removeErr != nil && !os.IsNotExist(removeErr) {
```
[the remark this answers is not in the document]
aneeshkp: I agree

pkg/daemon/daemon.go
```diff
@@ -293,13 +293,26 @@ func (dn *Daemon) Run() {
...
+							}
 						}
 					}
+					// Stop parent process last
 					p.cmdStop()
+					p.depProcess = nil
+					p.hasCollectedMetrics = false
+					//cleanup metrics
+					deleteMetrics(p.ifaces, p.haProfile, p.name, p.configName)
+					if p.name == syncEProcessName && p.syncERelations != nil {
```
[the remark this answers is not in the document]
aneeshkp: updated

josephdrichard (approved):

PR #65, opened 2025-07-09
Wait for cloud event proxy to be ready before starting linuxptp-daemon

nocturnalastro:


nocturnalastro:
Decided to take a different approach

PR #68, opened 2025-07-16
Allow configuring interval for enhanced log filtering

josephdrichard:
Also allow triggering when time exceeds threshold.

### DIFF
--- a/pkg/logfilter/logfilter.go
+++ b/pkg/logfilter/logfilter.go
@@ -2,6 +2,7 @@ package logfilter
 
 import (
 	"fmt"
+	"math"
 	"regexp"
 	"slices"
 	"strconv"
@@ -14,9 +15,10 @@ import (
 )
 
 const (
-	defaultEnhancedFilterTime = "30s"
-	ptp4lProcessName          = "ptp4l"
-	phc2sysProcessName        = "phc2sys"
+	defaultEnhancedFilterTime   = "30s"
+	defaultOffsetFlushThreshold = "0"
+	ptp4lProcessName            = "ptp4l"
+	phc2sysProcessName          = "phc2sys"
 )
 
 // LogFilter represents a specific filter that can be used to control logging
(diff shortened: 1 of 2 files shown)


pkg/logfilter/logfilter.go
```diff
@@ -188,9 +214,14 @@ func FilterOutput(logFilters []*LogFilter, output string) string {
 					continue
 				}
 			}
-			if filter.summaryText == "" {
+			if filter.enhancedThreshold > 0 && int64(math.Abs(filteredVal)) > filter.enhancedThreshold {
+				if len(filter.offsets) > 0 {
```
nocturnalastro, line 218: should this be greater than 1? It doesn't seem like there is much point if the only offset is the one we're about to output.

josephdrichard: If we trip based on the threshold, we never append the one that we're about to print.  So this would be the summary of all previous ones before the one that we're about to print.

pkg/logfilter/logfilter.go
```diff
@@ -99,13 +111,27 @@ func GetLogFilters(processName string, messageTag string, ptpSettings map[string
...
 	if filter, ok := ptpSettings["stdoutFilter"]; ok {
-		logFilters = append(logFilters, logFilterFromRegex(filter, nil, "", "", nil)) // Filter anything with specified filter
+		logFilters = append(logFilters, logFilterFromRegex(filter, nil, "", "", nil, defaultEnhancedFilterThreshold)) // Filter anything with specified filter
 	}
 	if logReduce, ok := ptpSettings["logReduce"]; ok {
-		if strings.ToLower(logReduce) == "true" || strings.ToLower(logReduce) == "basic" {
-			logFilters = append(logFilters, logFilterFromRegex("^.*master offset.*$", nil, "", "", nil)) // Just filter anything with master offset
-		} else if strings.ToLower(logReduce) == "enhanced" {
-			logFilters = getLogFiltersEnhanced(processName, messageTag, defaultEnhancedFilterTime)
+		logReduceSettings := strings.Split(strings.TrimSpace(logReduce), " ")
```
nocturnalastro, line 117: string.Fields will do the same without having to specify the space and will handle multiple spacing making it a bit more robust

pkg/logfilter/logfilter.go
```diff
@@ -14,9 +15,10 @@ import (
 )
 
 const (
-	defaultEnhancedFilterTime = "30s"
-	ptp4lProcessName          = "ptp4l"
-	phc2sysProcessName        = "phc2sys"
+	defaultEnhancedFilterTime      = "30s"
+	defaultEnhancedFilterThreshold = "0"
```
aneeshkp, line 19: isn';t this more of a offset related ?
could you name it as offsetFlushThreshold

josephdrichard: okay I can rename it

nocturnalastro (approved):
LGTM

PR #70, opened 2025-07-16
Add optional port name regex for ptp4l role events

nocturnalastro:


aneeshkp:
can  you add description please

pkg/parser/ptp4l_parser_internal_test.go
```diff
@@ -151,6 +151,18 @@ func TestPTP4LParser(t *testing.T) {
 				Event:      "UNCALIBRATED to LISTENING on RS_LISTENING",
 			},
 		},
+		{
+			name:    "Port state change to MASTER with port name",
+			logLine: "ptp4l[412707.219]: [ptp4l.0.config:5] port 11 (ens8f2): LISTENING to MASTER on ANNOUNCE_RECEIPT_TIMEOUT_EXPIRES",
```
aneeshkp, line 156: port name  is optional , we don't read port name instead we map port id with interface

aneeshkp: so older version doesn't depend on this - so we need to make sure we either ignore portname entirely or just read it and use it if available .

PR #82, opened 2025-08-21
Add alias manager

nocturnalastro:
use pci bus id with the last digit masked

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -1,7 +1,9 @@
 package utils
 
 import (
+	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


pkg/utils/utils.go
```diff
@@ -18,3 +23,66 @@ func GetAlias(ifname string) string {
...
+		}
+		for _, entry := range entries {
+			if strings.HasPrefix(entry.Name(), "lower_") {
+				path, err = FileSystem.Readlink(baseDir + "/" + entry.Name())
+				if err != nil {
+					glog.Errorf("failed to find pci bus ID for interface '%s': %s", ifname, err)
+					return ""
+				}
+				list := strings.Split(path, "/")
+				return list[len(list)-3]
```
vitus133, line 43: This could be worth protecting from panic if len(list) is smaller than 3. For example, if a virtual interface is picked, like here:
`lrwxrwxrwx.   1 root root    0 Sep  1 15:34 lower_00999b6dfa69205 -> ../00999b6dfa69205`
Also, what happens if there are several "lower_" ones under the iface? are they guaranteed to have the same PCI address?

PR #84, opened 2025-09-07
Send ptp4l event when running without proxy

vitus133:
This commit fixes T-BC ptp4l event when configured to run without the cloud-event proxy.

/cc @aneeshkp @jzding @josephdrichard @nocturnalastro


pkg/daemon/daemon.go
```diff
@@ -994,6 +994,9 @@ func (p *ptpProcess) cmdRun(stdoutToSocket bool, pm *PluginManager) {
 					printWhenNotEmpty(logfilter.FilterOutput(p.logFilters, output))
 					p.processPTPMetrics(output)
 					if p.name == ptp4lProcessName {
+						if profileClockType == TBC {
```
nocturnalastro, line 997: If this is for when not running cloud-event-proxy then should it have a check for the socket or stdoutToSocket?

josephdrichard (approved):

PR #86, opened 2025-09-10
Add chronyd support to daemon

josephdrichard:


josephdrichard:
closes: https://github.com/k8snetworkplumbingwg/ptp-operator/issues/81

PR #89, opened 2025-09-16
Suppress spurious log when GetAlias received "master" or an already aliased Intel or Mellanox interface                                                         

edcdavid:


nocturnalastro (approved):
/lgtm

PR #90, opened 2025-09-17
ntpfailover plugin implementation

josephdrichard:
Implement ntpfailover plugin to add support for failing over from GNSS to NTP as a time source for setting the system clock, in order to ensure at least some accuracy for the system clock even when GNSS is unavailable.

When this plugin is enabled, phc2sys will be disabled upon loss of GNSS, and reenabled once GNSS becomes available again.  Also, chronyd will be set to offline when GNSS is available, and then back online when GNSS stops being available.

Note that this is scoped to only allowing failover when GNSS is configured as the primary source for the system clock, and does not support failover from PTP to NTP.  Further work would be necessary to enable such a feature.

Also add required interfaces to plugin framework and throughout daemon code to support this implementation.


pkg/daemon/daemon.go
```diff
@@ -1170,8 +1196,8 @@ func (p *ptpProcess) cmdRun(stdoutToSocket bool, pm *PluginManager) {
 			break
 		} else {
 			glog.Infof("Recreating %s...", p.name)
-			newCmd := exec.Command(p.cmd.Args[0], p.cmd.Args[1:]...)
-			p.cmd = newCmd
+			newCmd := exec.Command(cmd.Args[0], cmd.Args[1:]...)
+			cmd = newCmd
```
nocturnalastro, line 1200: Shouldn't this be updating the struct?

pkg/daemon/daemon.go
```diff
@@ -1234,31 +1260,63 @@ func (p *ptpProcess) processPTPMetrics(output string) {
...
 	}
-	glog.Infof("removing config path %s for %s ", p.processConfigPath, p.name)
-	if p.processConfigPath != "" {
-		err := os.Remove(p.processConfigPath)
-		if err != nil {
-			glog.Errorf("failed to remove ptp4l config path %s: %v", p.processConfigPath, err)
+	<-p.exitCh
+}
+
+func (p *ptpProcess) cmdSetEnabled(pname string, enabled bool) {
```
nocturnalastro, line 1292: Does the process not know its name? Why is it passed as an argument?

pkg/daemon/daemon.go
```diff
@@ -1234,31 +1260,63 @@ func (p *ptpProcess) processPTPMetrics(output string) {
...
+	p.cmdSetEnabledMutex.Lock()
+	glog.Infof("cmdSetEnabled %s locked", p.name)
+	switch pname {
+	case "chronyd":
+		if enabled {
+			exec.Command("chronyc", "online").Output()
+		} else {
+			exec.Command("chronyc", "offline").Output()
+		}
+	case "phc2sys":
```
nocturnalastro, line 1303: should this just be for anything that isn't chronyd?

pkg/daemon/daemon.go
```diff
@@ -1234,31 +1260,63 @@ func (p *ptpProcess) processPTPMetrics(output string) {
...
+				glog.Infof("%s already running", p.name)
+			}
+		} else {
+			p.cmdStop()
 		}
 	}
-	<-p.exitCh
-	glog.Infof("Process %s (%d) terminated", p.name, p.cmd.Process.Pid)
+	glog.Infof("cmdSetEnabled %s unlocked", p.name)
+	p.cmdSetEnabledMutex.Unlock()
```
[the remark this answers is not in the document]
josephdrichard: what specifically are you referring to having it's own mutex?

pkg/daemon/daemon.go
nocturnalastro: shouldn't this be doing p.cmd = newCmd? 
Does the process struct get properly updated by doing this?

pkg/daemon/daemon.go
```diff
@@ -1234,31 +1260,63 @@ func (p *ptpProcess) processPTPMetrics(output string) {
...
+	case "chronyd":
+		if enabled {
+			exec.Command("chronyc", "online").Output()
+		} else {
+			exec.Command("chronyc", "offline").Output()
+		}
+	case "phc2sys":
+		if enabled {
+			if p.Stopped() && p.cmd != nil {
+				cmd := p.cmd
```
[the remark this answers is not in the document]
nocturnalastro: Sorry I can't put my comment against a block of code perhaps I was a bit vague. Could you just call p.Start() or p.cmdRun() or whatever the command is.

pkg/daemon/daemon.go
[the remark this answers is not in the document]
nocturnalastro: `cmdSetEnabledMutex` should there just mutext for protecting the process p.cmd changes as a whole instead its just protecting from multiple calls to this function. 


Also just realised this should probably be a defer from when you unlock otherwise it could end up in a bad state if we panic and it gets recovered at a higher level.

pkg/daemon/daemon.go
nocturnalastro: My worry is that if the process does die we will lose our handle on the running cmd and not being able to kill it when calling cmdStop.

pkg/daemon/daemon.go
```diff
@@ -1138,27 +1159,32 @@ func (p *ptpProcess) cmdRun(stdoutToSocket bool, pm *PluginManager) {
...
 		if !p.Stopped() {
-			err = p.cmd.Start() // this is asynchronous call,
+			glog.Infof("starting %s...", p.name)
+			p.cmd = cmd
+			err = cmd.Start() // this is asynchronous call,
 			if err != nil {
 				glog.Errorf("CmdRun() error starting %s: %v", p.name, err)
 			}
+
+			<-doneCh // goroutine is done
```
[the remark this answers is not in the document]
josephdrichard: We don't actually return if it errors out, just log, so it would still get to this point.

PR #91, opened 2025-09-25
Use GWH NIC's PCIAddress field.

greyerof:
The GWH package already provides a PCI Address for the NIC. Also, added some glog traces.


nocturnalastro (dismissed):
/lgtm

PR #182, opened 2026-03-23
Fix clock class reporting for dual NIC boundary clock

aabughosh:
## Summary

Fixes three compounding failures in the dual NIC BC scenario that caused cloud-event-proxy to never receive `CLOCK_CLASS_CHANGE` events:

- **Increase `clockClassRequestCh` buffer from 1 to 2**: When both NICs emit `CLOCK_CLASS_CHANGE` simultaneously, the channel buffer of 1 caused one request to be silently dropped. With dual NIC BC, both requests can now be queued.

- **Populate `clkSyncState` for BC clock class requests**: The `clockClassRequestCh` handler for BC type only updated the global `e.clockClass`/`e.clockAccuracy` but never wrote into the per-config `e.clkSyncState[cfgName]` map. The 60-second periodic ticker iterates `clkSyncState` to re-send clock class, so BC configs were never re-sent. This is the recovery mechanism when the initial write is lost to a broken pipe (cloud-event-proxy not ready on the Unix socket yet).

Together these bugs meant: the initial clock class write was lost to a broken pipe, the channel drop lost the retry for one NIC, and the ticker could never recover because `clkSyncState` was empty for BC configs.

## Test plan

- [ ] Deploy dual NIC BC configuration
- [ ] Verify both NICs' clock class changes are forwarded to cloud-event-proxy
- [ ] Verify the 60-second ticker re-sends clock class for BC configs after cloud-event-proxy restarts
- [ ] Confirm no `clock class request busy updating previous request` warnings in logs during simultaneous clock class changes

### DIFF
--- a/pkg/daemon/metrics.go
+++ b/pkg/daemon/metrics.go
@@ -593,6 +593,8 @@ func deleteMetrics(ifaces config.IFaces, haProfiles map[string][]string, process
 		return
 	}
 	deleteProcessStatusMetrics(config, process)
+	ClockClassMetrics.Delete(prometheus.Labels{
+		"process": ptp4lProcessName, "node": NodeName, "config": config})
 	for _, iface := range ifaces {
 		InterfaceRole.Delete(prometheus.Labels{
 			"process": ptp4lProcessName, "node": NodeName, "iface": iface.Name})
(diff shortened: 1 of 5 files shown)


sebsoto:
This may have been fixed by https://github.com/k8snetworkplumbingwg/linuxptp-daemon/pull/184

PR #187, opened 2026-03-31
OCPBUGS-79676: Fix hardware and ptp profile association

vitus133:
Align hardware config - ptp config association logic with the new naming scheme, where profile names are prefixed with resource name sepaated by "_".

/cc @nocturnalastro


nocturnalastro (approved):
/lgtm

PR #193, opened 2026-04-12
refactor: extract ptp4l config parsing into shared pkg/ptp4lconf package (CNF-20999)

gtannous-spec:
## Extract ptp4l config parsing into shared `pkg/ptp4lconf` package

- Extract ptp4l config parsing (types, `Populate`, `GetOption`, `SetOption`, clock-type inference) from `pkg/daemon/config.go` into a new shared `pkg/ptp4lconf` package. This eliminates the duplicated ad-hoc parser in `pkg/hardwareconfig/clockchain_resolution.go` and breaks the layering constraint that prevented `hardwareconfig` from reusing daemon's parser due to an import cycle. 
- Decouple clock type inference from ptp4l config parsing

**Motivation:** ptp4l config parsing was implemented independently in three places (daemon, hardwareconfig, and ptp-operator webhook) because the import graph `daemon -> hardwareconfig` prevented sharing. Extracting it to a leaf package lets both consume the same parser, reducing maintenance burden and behavioral inconsistency.
Addressing [CNF-20999](https://redhat.atlassian.net/browse/CNF-20999)

****_Assisted by Cursor_.****

### DIFF
--- a/pkg/daemon/config.go
+++ b/pkg/daemon/config.go
@@ -22,13 +22,6 @@ import (
 	ptpv1 "github.com/k8snetworkplumbingwg/ptp-operator/api/v1"
 )
 
-// Predefined config section names, re-exported from ptp4lconf for backward compatibility.
-const (
-	GlobalSectionName  = ptp4lconf.GlobalSectionName
-	NmeaSectionName    = ptp4lconf.NmeaSectionName
-	UnicastSectionName = ptp4lconf.UnicastSectionName
-)
-
 // LinuxPTPUpdate controls whether to update linuxPTP conf
 // and contains linuxPTP conf to be updated. It's rendered
 // and passed to linuxptp instance by daemon.
@@ -77,19 +70,19 @@ func (l *LinuxPTPConfUpdate) GetCurrentPTPProfiles() []string {
 	return profileNames
 }
 
-// Ptp4lConf wraps the shared ptp4lconf.Conf parser with daemon-specific
+// ProfileConfig wraps the shared ptp4lconf.Conf parser with daemon-specific
 // fields (profile name, GNSS serial port) and rendering methods.
-type Ptp4lConf struct {
+type ProfileConfig struct {
 	ptp4lconf.Conf
 	profileName    string
 	gnssSerialPort string
 }
 
-func (conf *Ptp4lConf) getPtp4lConfOptionOrEmptyString(sectionName, key string) (string, bool) {
+func (conf *ProfileConfig) getPtp4lConfOptionOrEmptyString(sectionName, key string) (string, bool) {
 	return conf.GetOption(sectionName, key)
 }
 
-func (conf *Ptp4lConf) setPtp4lConfOption(sectionName, key, value string, overwrite bool) {
+func (conf *ProfileConfig) setPtp4lConfOption(sectionName, key, value string, overwrite bool) {
 	conf.SetOption(sectionName, key, value, overwrite)
 }
 
@@ -176,29 +169,29 @@ func tryToLoadOldConfig(nodeProfilesJSON []byte) ([]ptpv1.PtpProfile, bool) {
 	return []ptpv1.PtpProfile{*ptpConfig}, true
 }
 
-// PopulatePtp4lConf delegates INI parsing and clock-type inference to ptp4lconf.Conf.Populate.
-func (conf *Ptp4lConf) PopulatePtp4lConf(config *string, cliArgs *string) error {
+// PopulatePtp4lConf takes as input a PtpProfile.ProfileConfig string and outputs as ProfileConfig struct
+func (conf *ProfileConfig) PopulatePtp4lConf(config *string, cliArgs *string) error {
 	return conf.Populate(config, cliArgs)
 }
 
 // ExtendGlobalSection extends Ptp4lConf struct with fields not from ptp4lConf
-func (conf *Ptp4lConf) ExtendGlobalSection(profileName string, messageTag string, socketPath string, pProcess string) {
+func (conf *ProfileConfig) ExtendGlobalSection(profileName string, messageTag string, socketPath string, pProcess string) {
 	conf.profileName = profileName
-	conf.setPtp4lConfOption(GlobalSectionName, "message_tag", messageTag, true)
+	conf.setPtp4lConfOption(ptp4lconf.GlobalSectionName, "message_tag", messageTag, true)
 	if socketPath != "" {
-		conf.setPtp4lConfOption(GlobalSectionName, "uds_address", socketPath, true)
+		conf.setPtp4lConfOption(ptp4lconf.GlobalSectionName, "uds_address", socketPath, true)
 	}
-	if gnssSerialPort, ok := conf.getPtp4lConfOptionOrEmptyString(GlobalSectionName, "ts2phc.nmea_serialport"); ok {
+	if gnssSerialPort, ok := conf.getPtp4lConfOptionOrEmptyString(ptp4lconf.GlobalSectionName, "ts2phc.nmea_serialport"); ok {
 		conf.gnssSerialPort = strings.TrimSpace(gnssSerialPort)
-		conf.setPtp4lConfOption(GlobalSectionName, "ts2phc.nmea_serialport", GPSPIPE_SERIALPORT, true)
+		conf.setPtp4lConfOption(ptp4lconf.GlobalSectionName, "ts2phc.nmea_serialport", GPSPIPE_SERIALPORT, true)
 	}
-	if _, ok := conf.getPtp4lConfOptionOrEmptyString(GlobalSectionName, "leapfile"); ok || pProcess == ts2phcProcessName { // not required to check process if leapfile is always included
-		conf.setPtp4lConfOption(GlobalSectionName, "leapfile", fmt.Sprintf("%s/%s", config.DefaultLeapConfigPath, os.Getenv("NODE_NAME")), true)
+	if _, ok := conf.getPtp4lConfOptionOrEmptyString(ptp4lconf.GlobalSectionName, "leapfile"); ok || pProcess == ts2phcProcessName { // not required to check process if leapfile is always included
+		conf.setPtp4lConfOption(ptp4lconf.GlobalSectionName, "leapfile", fmt.Sprintf("%s/%s", config.DefaultLeapConfigPath, os.Getenv("NODE_NAME")), true)
 	}
 }
 
-// AddInterfaceSection adds interface to Ptp4lConf
-func (conf *Ptp4lConf) AddInterfaceSection(iface string) {
+// AddInterfaceSection adds interface to ProfileConfig
+func (conf *ProfileConfig) AddInterfaceSection(iface string) {
 	ifaceSectionName := fmt.Sprintf("[%s]", iface)
 	conf.setPtp4lConfOption(ifaceSectionName, "", "", false)
 }
@@ -220,7 +213,7 @@ func getSource(isTs2phcMaster string) event.EventSource {
 //     (UNTIL next device section).
 //  2. Port section - any other section not starting with < (e.g. [eth0]) is the port section.
 //     Multiple port sections are allowed. Each port participates in SyncE communication.
-func (conf *Ptp4lConf) extractSynceRelations() *synce.Relations {
+func (conf *ProfileConfig) extractSynceRelations() *synce.Relations {
 	var err error
 	r := &synce.Relations{
 		Devices: []*synce.Config{},
@@ -267,7 +260,7 @@ func (conf *Ptp4lConf) extractSynceRelations() *synce.Relations {
 			synceRelationInfo.ExtendedTlv = extendedTlv
 		} else if strings.HasPrefix(sectionName, "[{") {
 			synceRelationInfo.ExternalSource = re.ReplaceAllString(sectionName, "")
-		} else if strings.HasPrefix(sectionName, "[") && sectionName != GlobalSectionName {
+		} else if strings.HasPrefix(sectionName, "[") && sectionName != ptp4lconf.GlobalSectionName {
 			iface := re.ReplaceAllString(sectionName, "")
 			ifaces = append(ifaces, iface)
 		}
@@ -282,14 +275,14 @@ func (conf *Ptp4lConf) extractSynceRelations() *synce.Relations {
 }
 
 // RenderSyncE4lConf outputs synce4l config as string
-func (conf *Ptp4lConf) RenderSyncE4lConf(ptpSettings map[string]string) (configOut string, relations *synce.Relations) {
-	configOut = fmt.Sprintf("#profile: %s\n", conf.profileName)
+func (conf *ProfileConfig) RenderSyncE4lConf(ptpSettings map[string]string) (configOut string, relations *synce.Relations) {
+	configOut = "#profile: " + conf.profileName + "\n"
 	relations = conf.extractSynceRelations()
 	relations.AddClockIds(ptpSettings)
 	deviceIdx := 0
 
 	for i, section := range conf.Sections {
-		configOut = fmt.Sprintf("%s\n%s", configOut, section.SectionName)
+		configOut += "\n" + section.SectionName
 		if strings.HasPrefix(section.SectionName, "[<") {
 			if _, found := conf.getPtp4lConfOptionOrEmptyString(section.SectionName, "clock_id"); !found {
 				conf.setPtp4lConfOption(section.SectionName, "clock_id", relations.Devices[deviceIdx].ClockId, true)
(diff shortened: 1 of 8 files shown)


pkg/ptp4lconf/ptp4lconf.go
```diff
@@ -0,0 +1,170 @@
...
+				if len(currentLine) < 2 {
+					return errors.New("Section missing closing ']': " + line)
+				}
+				currentSectionName = fmt.Sprintf("%s]", currentLine[0])
+				if currentSectionName != GlobalSectionName && currentSectionName != NmeaSectionName && currentSectionName != UnicastSectionName {
+					ifaceCount++
+				}
+				c.SetOption(currentSectionName, "", "", false)
+			} else {
+				split := strings.IndexByte(line, ' ')
```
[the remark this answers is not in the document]
gtannous-spec: That was a pre-existing logic, The only change was refactoring the architecture of these packages. 
However the convention goes with a single separator in the ptp4l configuration between the key & value. 

WDYT @vitus133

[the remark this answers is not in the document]
nocturnalastro: I see no reason not to add that. ptpconfig takes the config file as user input so we should be defensive around parsing it.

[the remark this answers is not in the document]
nocturnalastro: Add changes like this into a seperate commit though.

pkg/ptp4lconf/ptp4lconf.go
```diff
@@ -0,0 +1,170 @@
...
+	name = strings.ReplaceAll(name, "[", "")
+	name = strings.ReplaceAll(name, "]", "")
+	return name
+}
+
+// RenderOptions renders all options in a section as newline-separated "key value" lines.
+func RenderOptions(s Section) string {
+	var out string
+	for _, opt := range s.Options {
+		out = fmt.Sprintf("%s\n%s %s", out, opt.Key, opt.Value)
```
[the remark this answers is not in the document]
nocturnalastro: I think this is much easier to read. 
```suggestion
}

// RenderOptions renders all options in a section as newline-separated "key value" lines.
func RenderOptions(s Section) string {
	var out string
	for _, opt := range s.Options {
		out += "\n" + opt.Key + " " + opt.Value
```

pkg/ptp4lconf/ptp4lconf.go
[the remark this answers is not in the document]
gtannous-spec: I agree it would cover more edge cases.
However, I just didn't want to touch and change the logic of the parser functionality that existed before. 
if that is a good change in ur opinion, I can commit this suggestion.

pkg/ptp4lconf/ptp4lconf.go
```diff
@@ -0,0 +1,170 @@
...
+						(key == "slaveOnly" && value == "1") ||
+						(key == "clientOnly" && value == "1") {
+						hasSlaveConfigDefined = true
+					}
+				}
+			}
+		}
+	}
+
+	if !hasSlaveConfigDefined {
```
nocturnalastro: I'd like to factor out the clock type logic. But perhaps we change that that later. We should do something like store the number of ports with `masterOnly|serverOnly` true, false or default, and the same for `slaveOnly|clientOnly`. 
Doing that in the parsing step would allow us to later calculate the clock type. And would mean we wouldn't have to do strange things like pass the cli flags into here.

pkg/daemon/config.go
```diff
@@ -404,68 +283,57 @@ func (conf *Ptp4lConf) extractSynceRelations() *synce.Relations {
...
 				deviceIdx++
 			}
 		}
-		for _, option := range conf.sections[i].options {
-			k := option.key
-			v := option.value
-			configOut = fmt.Sprintf("%s\n%s %s", configOut, k, v)
+		for _, option := range conf.Sections[i].Options {
+			configOut = fmt.Sprintf("%s\n%s %s", configOut, option.Key, option.Value)
 		}
```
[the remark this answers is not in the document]
gtannous-spec: @nocturnalastro WDYT

pkg/daemon/config.go
```diff
@@ -22,16 +22,13 @@ import (
 	ptpv1 "github.com/k8snetworkplumbingwg/ptp-operator/api/v1"
 )
 
-// predefined config section names
+// Predefined config section names, re-exported from ptp4lconf for backward compatibility.
```
nocturnalastro: This kind of thing creates a long tail of things to fix. Not sure if we should be doing that.

pkg/daemon/config.go
```diff
@@ -80,75 +77,20 @@ func (l *LinuxPTPConfUpdate) GetCurrentPTPProfiles() []string {
...
-type ptp4lConfSection struct {
-	sectionName string
-	options     []ptp4lConfOption
-}
-
-// Ptp4lConf is a structure to represent a parsed ptpconfig,
-// which can then be rendered to a string again.
+// Ptp4lConf wraps the shared ptp4lconf.Conf parser with daemon-specific
+// fields (profile name, GNSS serial port) and rendering methods.
 type Ptp4lConf struct {
```
nocturnalastro: I feel like this should be something like PtpProfile or something like that. As it holds the profile info. Might help with confusing the two.

nocturnalastro: Clock type should also probably live here but ATM its so interlaced with the parsing code I see why it isn't

pkg/daemon/config.go
gtannous-spec: Revert changes or fully remove the comment? 
Btw removing comment causes to linter error

pkg/daemon/config.go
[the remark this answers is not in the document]
nocturnalastro: Same thing as the other one isn't it?

pkg/ptp4lconf/ptp4lconf.go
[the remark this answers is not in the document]
nocturnalastro: any improvements you make just make them in a separate commit and make sure to test them.

pkg/daemon/config.go
nocturnalastro: Yeah there should be a newline before the const block otherwise it treats it as a doc string I think. 

My Comment is more on about the approach. The compiler/lsp will complain if you miss a reference so you can be pretty confident you'll catch them all, so I see no need to re-export. Its just yet another indirection and something we need to clean up later.

pkg/daemon/config.go
```diff
@@ -234,83 +176,20 @@ func tryToLoadOldConfig(nodeProfilesJSON []byte) ([]ptpv1.PtpProfile, bool) {
 	return []ptpv1.PtpProfile{*ptpConfig}, true
 }
 
-// PopulatePtp4lConf takes as input a PtpProfile.Ptp4lConf string and outputs as ptp4lConf struct
+// PopulatePtp4lConf delegates INI parsing and clock-type inference to ptp4lconf.Conf.Populate.
```
vitus133: Why do we care what it delegates where? I'd rather read what the function inputs, outputs and what it does. Should we remove this function and replace with the new one at all callers?

pkg/daemon/config.go
gtannous-spec: `ProfileConfig` would make more sense than Ptp4lConf?
WDYT @nocturnalastro

pkg/daemon/config.go
gtannous-spec: do you mean we have to get rid of these consts or what (and just define them inplace where they are called)

pkg/daemon/config.go
nocturnalastro: `ProfileConfig` sounds too generic. `PtpProfile` at least tells you somewhat what its for. perhaps `Ptp4lProfile`?

@vitus133 Am I bikeshed-ing here?

pkg/daemon/config.go
nocturnalastro: They should be used from the new package where they originate.

sebsoto: Redefining these to the same values as the other package adds a weird layer of indirection/obfuscation.

pkg/daemon/config.go
nocturnalastro: You're embedding the config so it will automatically call it there no need to delegate.

pkg/ptp4lconf/ptp4lconf.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// Section is a named group of options in a ptp4l INI-style config (e.g. "[global]").
+type Section struct {
+	SectionName string
+	Options     []Option
+}
+
+// Conf represents a parsed ptp4l INI-style configuration
+// that can be queried, mutated, and rendered back to text.
```
sebsoto: Nit: This line feels like fluff

pkg/ptp4lconf/ptp4lconf.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	if !found {
+		updatedSection.Options = append(updatedSection.Options, Option{Key: key, Value: value})
+	}
+	c.Sections[index] = updatedSection
+}
+
+// Populate parses a ptp4l INI-style config string and optional CLI args,
+// populating the Conf struct with sections, options, and inferred clock type.
+func (c *Conf) Populate(config *string, cliArgs *string) error {
```
sebsoto: I would encourage changing this to `func Parse(config, cliArgs *string) (*Conf,error)`.
It is a more ergonomic and idiomatic pattern in go.


Looking at how this function is called, it is always done after declaring `conf := &Ptp4lConf{}`. 
There does not seem to be a reason to require calling c.Populate on an existing Conf object. If there ends up being one, that special case can be handled easily.

pkg/ptp4lconf/ptp4lconf.go
```diff
@@ -0,0 +1,170 @@
...
+// SectionName strips bracket characters from a section name header,
+// returning the bare name (e.g. "[ens1f0]" -> "ens1f0").
+func SectionName(name string) string {
+	name = strings.ReplaceAll(name, "[", "")
+	name = strings.ReplaceAll(name, "]", "")
+	return name
+}
+
+// RenderOptions renders all options in a section as newline-separated "key value" lines.
+func RenderOptions(s Section) string {
```
sebsoto: This should be a method of `Section`

pkg/ptp4lconf/ptp4lconf.go
```diff
@@ -0,0 +1,170 @@
...
+	} else {
+		c.ClockType = event.OC
+	}
+
+	return nil
+}
+
+// SectionName strips bracket characters from a section name header,
+// returning the bare name (e.g. "[ens1f0]" -> "ens1f0").
+func SectionName(name string) string {
```
sebsoto: This should be a `Name()` method of `Section`

sebsoto: SectionName only ever operates on a Section's own field, so it should be a method on the type. It helps with code organization and ease of use for someone using the library.

sebsoto (changes requested):

pkg/daemon/config.go
gtannous-spec: The webhook parser has these "original" names too, this might be a convention between the ptp-operator webhook svc account and the current parser logic. 

Making such changes like struct/function names level and arguments can lead to confusion, though changing it here could be easy.

nocturnalastro: ptp-operator imports from linuxptp-daemon? Interesting. I think this is the time to change if we were going to.

pkg/ptp4lconf/ptp4lconf.go
nocturnalastro: Where is this used?

pkg/ptp4lconf/ptp4lconf.go
nocturnalastro: Why isn't this just a feild? We could strip the brackets off in the parsing step.

nocturnalastro: Or a key in a hashmap?

pkg/daemon/config.go
gtannous-spec: not really imports :P just adopts the naming convention of those parsers in lptpd

nocturnalastro: Oh ok well this PR's moved them, then ptp-operator can probably should just import it :)

pkg/ptp4lconf/ptp4lconf.go
nocturnalastro: Also its not strictly INI-style as INI has `key=value` not `key value [values..]` Its why we don't just use an INI parser.

gtannous-spec: Fixed that

pkg/ptp4lconf/ptp4lconf.go
gtannous-spec: Yeah that's a dead code cursor generated after creating a new ptp4lconf pkg, should remove it. 
Thanks.

pkg/daemon/config.go
```diff
@@ -80,75 +70,20 @@ func (l *LinuxPTPConfUpdate) GetCurrentPTPProfiles() []string {
...
 
-// Ptp4lConf is a structure to represent a parsed ptpconfig,
-// which can then be rendered to a string again.
-type Ptp4lConf struct {
-	sections         []ptp4lConfSection
-	profile_name     string
-	clock_type       event.ClockType
-	gnss_serial_port string // gnss serial port
+func (conf *ProfileConfig) getPtp4lConfOptionOrEmptyString(sectionName, key string) (string, bool) {
+	return conf.GetOption(sectionName, key)
```
nocturnalastro: Why do these need to exist?

pkg/daemon/config.go
```diff
@@ -80,75 +70,20 @@ func (l *LinuxPTPConfUpdate) GetCurrentPTPProfiles() []string {
...
-	}
-	// Append unless already overwrote it.
-	if !found {
-		updatedSection.options = append(updatedSection.options, ptp4lConfOption{key: key, value: value})
-	}
-
-	//Update section in conf
-	conf.sections[index] = updatedSection
+func (conf *ProfileConfig) setPtp4lConfOption(sectionName, key, value string, overwrite bool) {
+	conf.SetOption(sectionName, key, value, overwrite)
```
nocturnalastro: Why do these need to exist?

pkg/daemon/config.go
```diff
@@ -234,92 +169,29 @@ func tryToLoadOldConfig(nodeProfilesJSON []byte) ([]ptpv1.PtpProfile, bool) {
...
-		conf.clock_type = event.BC
-	} else {
-		// Single slave Interface defined
-		conf.clock_type = event.OC
-	}
-
-	return nil
+// PopulatePtp4lConf takes as input a PtpProfile.ProfileConfig string and outputs as ProfileConfig struct
+func (conf *ProfileConfig) PopulatePtp4lConf(config *string, cliArgs *string) error {
+	return conf.Populate(config, cliArgs)
```
nocturnalastro: Why do these need to exist?

pkg/ptp4lconf/ptp4lconf.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	ifaceCount := 0
+	if config != nil {
+		for _, line := range strings.Split(*config, "\n") {
+			line = strings.TrimSpace(line)
+			if line == "" || strings.HasPrefix(line, "#") {
+				continue
+			} else if strings.HasPrefix(line, "[") {
+				currentLine := strings.Split(line, "]")
```
nocturnalastro: Or just use HasSuffix, then you don't have to manipulate the string on line 121.

pkg/ptp4lconf/ptp4lconf_test.go
```diff
@@ -0,0 +1,210 @@
+package ptp4lconf
+
+import (
+	"testing"
+
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/event"
+	"github.com/stretchr/testify/assert"
+)
+
+func stringPtr(s string) *string { return &s }
```
nocturnalastro: https://pkg.go.dev/k8s.io/utils/ptr#To

nocturnalastro (changes requested):

pkg/daemon/config.go
gtannous-spec: You're right we can get rid of those.

pkg/daemon/config.go
gtannous-spec: Will get rid of this as well

pkg/ptp4lconf/ptp4lconf.go
gtannous-spec: > I would encourage changing this to func Parse(config, cliArgs *string) (*Conf,error).
It is a more ergonomic and idiomatic pattern in go.

I kinda agree on that one, but I prefer to leave it as it is for now since it was previously called like that, it's called the same thing too in the webhook of the ptp-operator where the ptpconfig is parsed for validation purposes, but I agree that it would be way more readable once we change it to parse instead of populate.

> Looking at how this function is called, it is always done after declaring conf := &Ptp4lConf{}.
There does not seem to be a reason to require calling c.Populate on an existing Conf object. If there ends up being one, that special case can be handled easily.

Same as above, if you think this step would be more crucial for future code maintenance I would suggest a change on this case and (probably other similar cases if applicable) on a different PR/commit.

WDYT @sebsoto

pkg/ptp4lconf/ptp4lconf.go
gtannous-spec: Changed to be a field of Section struct

pkg/ptp4lconf/ptp4lconf.go
nocturnalastro: @gtannous-spec Can you write a tech debt for doing this at some point so it doesn't get lost.

gtannous-spec: Done it (used the current parsing method now, not the one from this PR but should be clear tho) 

[CNF-23109](https://redhat.atlassian.net/browse/CNF-23109) WDYT @nocturnalastro

gtannous-spec:
Will open an enhanced version of this PR later !

PR #195, opened 2026-04-18
001 process failure handling - requirements and assumptions

vitus133:
This is a requirements and assumptions phase for process failure handling mechanism, initially discussed in https://docs.google.com/document/d/1olV_MC7hW3vj5CZY-2KbAeMsbu_Ukq6Ru_eu8j7cVJU/edit?usp=sharing

Most of the files here are standard templates of speckit and its cursor plugins, and are not directly related to our work. The files that worth your attention are:

[specs/001-process-failure-handling/checklists/requirements.md](https://github.com/k8snetworkplumbingwg/linuxptp-daemon/compare/main...vitus133:001-process-failure-handling?expand=1#diff-aa2fa3a79ed4c7a95cd62c4f8acfbc7645e44e389fe46167df5d816a407639a3)

[specs/001-process-failure-handling/current-behavior.md](https://github.com/k8snetworkplumbingwg/linuxptp-daemon/compare/main...vitus133:001-process-failure-handling?expand=1#diff-2824f30b63b3526cf666c8525957febdd659e46be9348551197f9a0fbace4077)

[specs/001-process-failure-handling/spec.md](https://github.com/k8snetworkplumbingwg/linuxptp-daemon/compare/main...vitus133:001-process-failure-handling?expand=1#diff-14ff82120dfd3970ce7a6d0c32231ce8f898bd014f9e4363490da518a18a39ed)

/cc @nocturnalastro @lack @jzding @sebsoto

### DIFF
--- a/specs/001-process-failure-handling/spec.md
+++ b/specs/001-process-failure-handling/spec.md
@@ -190,14 +190,18 @@ works (CNF-21499 fixed).
 
 When processes fail in a T-GM deployment, the system MUST handle each
 process type appropriately:
-- **ts2phc failure**: Don't trigger any events if recovered within the downtime threshold.
-  If not recovered within the downtime threshold, the system MUST emit the appropriate state-change events.
+- **ts2phc failure** (**holdover-breaking** while down past the threshold):
+  Per FR-009 / FR-013, do not downgrade state or class until the downtime
+  threshold expires; then FREERUN and class 248. If recovered within the
+  threshold, suppress the suppressed event categories (FR-017).
 - **ptp4l failure**: Since T-GM ptp4l is a time transmitter, downstream
   clients lose their master. The system MUST announce freerun clock class
   until ptp4l recovers.
-- **gpsd/gpspipe failure**: If GNSS signal is already acquired and DPLL
-  maintains lock, short outages SHOULD be tolerated via the downtime
-  threshold. Extended outages MUST trigger GNSS-lost state transitions.
+- **gpsd/gpspipe failure** (**holdover-capable** when ts2phc keeps running
+  and the GM can remain in true holdover): Short outages SHOULD be
+  tolerated via the downtime threshold (no premature class or state
+  changes). Extended outages MUST trigger GNSS-lost state transitions and
+  announcements consistent with actual timing state (FR-013).
 
 **Why this priority**: T-GM is the timing root; incorrect behavior here
 propagates errors to all downstream devices in the network.
@@ -277,6 +281,8 @@ announcements per the acceptance scenarios.
 - Q: Are synce4l and chronyd failures in scope? → A: Out of scope. They operate independently (SyncE is physical-layer frequency, chronyd is optional NTP bridge) and their failures don't interact with the PTP clock state machine.
 - Q: Is OC and BC (non-telecom) failure handling in scope? → A: Deferred. FR-014 is retained as a noted requirement but implementation is deferred to a follow-up feature. T-BC/T-GM modes are the critical path.
 - Q: What GM state should the system enter when ts2phc fails past the downtime threshold — FREERUN or HOLDOVER (following DPLL)? → A: Always FREERUN. The system cannot maintain holdover with ts2phc down; the timing chain from GNSS/1PPS to PHC is broken regardless of DPLL hardware state.
+- Q: How do FR-009 and FR-013 relate for T-GM (clock class 7 vs FREERUN)? → A: Class 7 applies only when the GM is genuinely in holdover (outages that do not break the ts2phc-mediated timing path). ts2phc loss past the threshold is holdover-breaking: state FREERUN and announced class 248, not 7.
+- Q: May clock state or class change before the downtime threshold expires for threshold-governed failures? → A: No. Until the threshold expires (per FR-004), clock state, PMC clock class, and suppressed synchronization/GM-status events must not change solely because of that outage. Immediate transitions required elsewhere (e.g. FR-006) are unchanged.
 - Q: How does TT ptp4l failure get communicated to downstream clients? → A: Daemon-side only. The daemon emits log lines and metric updates for the cloud-event-proxy and monitoring systems. Downstream PTP clients detect the loss via ANNOUNCE_RECEIPT_TIMEOUT. No pre-emptive PMC push to a dead TT ptp4l is needed.
 
 ## Requirements *(mandatory)*
@@ -312,10 +318,19 @@ announcements per the acceptance scenarios.
   for `CLOCK_REALTIME`. These events are only appropriate for phc2sys
   failures affecting system clock synchronization.
 
-- **FR-009**: In T-GM mode, ts2phc process death past the downtime
-  threshold MUST transition GM state to FREERUN. The system cannot
-  maintain holdover with ts2phc down because the timing chain from
-  GNSS/1PPS to PHC is broken regardless of DPLL hardware state.
+- **FR-009**: In T-GM mode, ts2phc process death is **holdover-breaking**:
+  the timing chain from GNSS/1PPS through ts2phc to the PHC is not
+  credible while ts2phc is down, regardless of DPLL hardware indications.
+  **Until** the configured downtime threshold expires without stable
+  recovery (FR-004), the daemon MUST NOT transition GM timing state to
+  FREERUN, MUST NOT announce freerun clock class for this cause, and MUST
+  inhibit the clock-state, clock-class, and synchronization/GM-status
+  events that the threshold suppresses (FR-017). **After** the threshold
+  expires without recovery, GM timing state MUST transition to FREERUN
+  and announcements MUST match that state (FR-013 holdover-breaking case;
+  class 248). This is mandatory: downstream clients must not observe a
+  holdover class for a broken GNSS-to-PHC path once the outage is treated
+  as real.
 
 - **FR-010**: In T-GM mode, ptp4l (time transmitter) process death MUST
   announce freerun clock class to downstream clients.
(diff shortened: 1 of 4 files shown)


specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+immediately — identical to how an upstream link-down is handled today. If
+the system was in any other state (FREERUN, NOTSET), it MUST remain in or
+transition to FREERUN. Upon ptp4l recovery and PTP adjacency
+re-establishment, the system follows the normal path: determining sufficient
+synchronization, locking DPLL on the PHC, and entering LOCKED state.
+Currently, the daemon emits `os-clock-sync-state-change` and
+`synchronization-state-change` events for `CLOCK_REALTIME` (the same events
+phc2sys emits), which is incorrect: the outage affects the upstream PTP
+source, not the system clock synchronization.
+
```
nocturnalastro, line 24: I'd like a mermaid/ascii diagram here  if these are meant to be human and AI readable I think it would help.

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+   with clock class 248.
+
+3. **Given** a T-BC or T-TSC deployment in HOLDOVER due to TR ptp4l
+   restart,
+   **When** the TR ptp4l process recovers, PTP adjacency is
+   re-established, sufficient synchronization is determined, and DPLL
+   locks on the PHC,
+   **Then** the BC state machine transitions to LOCKED, clock class reverts
+   to the upstream GM class, and a clock class change event is emitted.
+
```
nocturnalastro, line 58: If the restart is within a small enough timeframe should we even change our clock class?

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+events are emitted at all, leaving monitoring consumers unaware of the
+outage.
+
+**Why this priority**: TT ptp4l failure directly affects all downstream PTP
+clients. Silent failures violate the observability principle and can cause
+downstream clock drift without operator awareness.
+
+**Independent Test**: Kill the TT ptp4l process in a T-BC deployment, verify
+a FREERUN event is emitted, and verify the system returns to LOCKED after
+recovery. Verify clock class reflects the FREERUN state during the outage.
```
nocturnalastro, line 89: Shouldn't the downstream clocks be the ones notifying of the failure here?

specs/001-process-failure-handling/spec.md
nocturnalastro: NVM looks like this is handled below

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+   preserved).
+
+4. **Given** a per-process downtime threshold configuration,
+   **When** different thresholds are set for ptp4l (0s), phc2sys (5s), and
+   ts2phc (3s),
+   **Then** each process respects its own threshold independently.
+
+---
+
+### User Story 4 - Short-Lived phc2sys/ts2phc Restarts Are Silenced (Priority: P2)
```
nocturnalastro, line 151: Don't the **Acceptance Scenarios** for `US3` require this?

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+   emitted.
+
+3. **Given** a T-GM deployment with ts2phc killed,
+   **When** ts2phc fails to restart within the threshold,
+   **Then** a GM status change event to FREERUN is emitted. The system
+   cannot maintain holdover with ts2phc down.
+
+---
+
+### User Story 5 - T-GM Process Failure Behavior (Priority: P2)
```
nocturnalastro, line 189: I feel like this should also be coverted by US3. I feel like are just details of US3

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+
+### Edge Cases
+
+- What happens when multiple processes fail simultaneously (e.g., ptp4l and
+  phc2sys both killed)?
+  The system MUST handle each failure individually and emit the appropriate state-change events.
+- How does the system behave when a process enters a crash loop (repeated
+  failures within seconds)?
+  The downtime threshold timer starts on the first crash and does not reset
+  on subsequent crashes. The process must be up for 2 consecutive seconds
```
nocturnalastro, line 250: I would say it should reach something like emitting some form of state log line but perhaps it being time-based is an ok first step.

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+  determines which processes run and how failures are interpreted.
+
+## Success Criteria *(mandatory)*
+
+### Measurable Outcomes
+
+- **SC-001**: In T-BC / T-TSC mode, TR ptp4l process death while in
+  LOCKED state results in holdover entry 100% of the time, with clock
+  class 135 announced within 2 seconds. TR ptp4l death while in
+  FREERUN/NOTSET results in FREERUN with clock class 248.
```
nocturnalastro, line 425: Unless it recovers within the defined process threshold? Otherwise what is the point in US3+

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+
+- **SC-002**: In T-BC mode, TT ptp4l process death results in freerun
+  announcement 100% of the time, with recovery to locked within 5 seconds
```
nocturnalastro, line 428: Same as above

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+  of ports returning to MASTER state.
+
+- **SC-003**: When configured with a downtime threshold, phc2sys and
+  ts2phc restarts within the threshold produce zero spurious state-change
```
nocturnalastro, line 432: We have to define spurious in with given that SC-001 and SC-002 say they should happen 100% of the time.

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+  events.
+
+- **SC-004**: ts2phc process restarts recover successfully 100% of the
+  time (CNF-21499 resolved), with the process functional within 3 seconds.
```
nocturnalastro, line 436: Assuming there is no other reason for it to crash loop like a bad config.

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+
+- **SC-005**: In T-GM mode, ts2phc failure past the downtime threshold
+  results in FREERUN state 100% of the time. ptp4l (TT) failure results
```
nocturnalastro: I would be prefer something like `MUST result in` than `100% of the time`

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+  in freerun clock class (248) announcement within 2 seconds.
+
+- **SC-006**: All process failures across all clock modes (T-BC, T-GM,
+  OC, BC) produce `PTP_PROCESS_STATUS` log lines within 1 second.
+
+- **SC-007**: No false CLOCK_REALTIME synchronization events are emitted
+  when the actual failure is in the upstream PTP source (TR ptp4l).
```
nocturnalastro, line 447: Do we define when it should happen otherwise how do we know its false?

False like spurious above it assume knowledge of when it should happen. I don't we should assume that without a reference to something else where that can be found.

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+
+- **SC-008**: The system handles 10 consecutive process crash/restart
```
nocturnalastro, line 449: At least 10?

specs/001-process-failure-handling/spec.md
```diff
@@ -0,0 +1,413 @@
...
+- **FR-011**: The ts2phc process MUST restart and become functional within
+  3 seconds of process death, 100% of the time (addresses CNF-21499).
+
+- **FR-012**: The `Reset` event sent on ts2phc death MUST NOT clear
+  state that is still valid (e.g., DPLL state, GNSS fix status) while
+  the process is within its configured downtime threshold window.
+
+- **FR-013**: Clock class change announcements via PMC MUST reflect the
+  actual timing state during process outages. In holdover, the announced
+  class MUST be 135 (in-spec) or 165 (out-of-spec) for T-BC / T-TSC. For T-GM, the class MUST be 7 (holdover).
```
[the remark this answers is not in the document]
nocturnalastro: Until the timeout threshold has been reached?

sebsoto:
I don't use cursor so I cant really review the cursor specific files. I'm also not familiar with spec-kit, so looking into that. Here's some general thoughts:

I am a bit worried about putting LLM brand specific files in the top level directory of the repo. It feels similar to checking in `.jetbrains/`, `.eclipse/`, etc. 

The use of `AGENTS.md` is common among projects in order to prevent `CURSOR.md`, `CLAUDE.md`, from both polluting the repo, but also to reduce the maintenance burden caused by the duplication that may occur between them. It looks like `.agents/skills/` might be the place for stuff to go.

Are we 100% sure that we want to check skills into this repo? Are we requiring that contributors purchase $MODEL licenses to contribute? Is $MODEL going to be available for the knowable future? What happens if our main contributors lose access? What if a change $MODEL deteriorates its ability and breaks these skills? This is closed source, proprietary software that is outside of our control.

If some of these skills are not specific to our product, do they really belong in the repo?
If it is helpful to share generic $MODEL skills with others, could a secondary repo within the org be of use?

To be clear, I am not completely opposed to this, just want to make sure we work everything out before making a decision to commit to it :smile:

specs/001-process-failure-handling/spec.md
vitus133: Downstream clocks will catch the failure, but I wonder whether we should change our metrics on the monitoring level. Maybe we should leave this decision to user and allow configuring "infinity" as the downtime threshold?

nocturnalastro: so we'd need to have a config per ptp profile. As we might want config different for TR and TT

vitus133:
Closing as it will be covered by the spec-driven requirements project

PR #197, opened 2026-04-24
add coverage gate to block PRs with decreasing coverage

jzding:
Add a GitHub Action to enforce code coverage standards.

- Executes tests with SKIP_GNSS_MONITORING=1 and --tags=unittests.
- Filters out auto-generated Kubernetes code (zz_generated),
  protobuf (.pb.go), and mocks to ensure high-signal metrics.
- On PRs, generates baseline coverage from the base branch and
  blocks if coverage decreases relative to that baseline.
- Supports skip-coverage label to bypass the check when needed.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -19,21 +19,10 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Coverage Gate
+        if: github.event_name == 'pull_request'
+        run: make coverage-gate BASE_REF=${{ github.event.pull_request.base.sha }}
+
       - name: Generate Coverage
+        if: github.event_name == 'push'
         run: make test
-
-      - name: Generate Base Coverage
-        if: github.event_name == 'pull_request'
-        run: |
-          git checkout ${{ github.event.pull_request.base.sha }}
-          make test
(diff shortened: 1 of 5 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,39 @@
...
+          go run github.com/vladopajic/go-test-coverage/v2@v2.17.1 -p coverage.out --breakdown-file-name base-coverage.json
+          git checkout ${{ github.sha }}
+
+      - name: Check Coverage Regressions
+        if: github.event_name == 'pull_request'
+        uses: vladopajic/go-test-coverage@v2
+        with:
+          profile: coverage.out
+          base-breakdown-file: base-coverage.json
+          diff-threshold: 0
```
edcdavid: looks good to me, it would be nice to have a way to locally compute the diff before pushing a PR (could be done later in another PR)

sebsoto:
We had discussed in the upstream meeting two weeks ago that we would like to start with informing on test coverage change for PRs at first, rather than enforcing coverage. Maybe we should start with just the inform part?

.github/workflows/coverage.yml
jzding: added script so that users can run it locally to check the coverage gate result.
`make coverage-gate`

edcdavid (approved):

PR #199, opened 2026-04-24
[WIP] Do Not Merge: Remove stdoutToSocket flag

sebsoto:
The daemon previously supported two modes: with or without the cloud-event-proxy sidecar, controlled by the LOGS_TO_SOCKET env var and stdoutToSocket flag. In practice, cloud-event-proxy is always deployed, resulting in the maintenance of two major deployment methods for little gain. By removing the flag, the code execution flow is cleaned up, and the code as a whole becomes easier to parse.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
-	"strconv"
 	"strings"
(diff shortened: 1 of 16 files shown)


sebsoto:
Just seeing what this would look like + checking if tests pass

PR #203, opened 2026-05-10
Always update Prometheus clock_class metric regardless of socket mode

aabughosh:
## Summary

When cloud-event-proxy is deployed (stdoutToSocket=true), the openshift_ptp_clock_class Prometheus metric is never updated. Both emitClockClass() and UpdateClockClass() guard the metric update behind !stdoutToSocket, so the metric remains stale while the actual clock class changes.

This breaks any monitoring or conformance tests that rely on the Prometheus endpoint to observe clock class transitions (e.g. LOCKED to FREERUN after port down).

**Root cause:** The else-if / if !stdoutToSocket condition treats socket writes and Prometheus metric updates as mutually exclusive, but they serve different consumers and should both be active.

**Fix:** Decouple the Prometheus metric update from the socket write condition. The metric is now always set when clockClassMetric is non-nil, regardless of whether events are also forwarded to cloud-event-proxy.

## Changes

- emitClockClass(): Changed if !e.stdoutToSocket && e.clockClassMetric != nil to if e.clockClassMetric != nil
- UpdateClockClass(): Changed else if e.clockClassMetric != nil to separate if e.clockClassMetric != nil block
- Added TestEmitClockClass_UpdatesPrometheusMetricWithSocket test verifying that the Prometheus metric is updated even when stdoutToSocket is true

## Test plan

- [ ] New unit test TestEmitClockClass_UpdatesPrometheusMetricWithSocket passes
- [ ] Existing TestEmitClockClass_* tests still pass
- [ ] E2E: Verify clockClass when locking PTP source on single NIC boundary clock test passes with cloud-event-proxy deployed

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -657,7 +657,7 @@ func (e *EventHandler) emitClockClass(clockClass fbprotocol.ClockClass, cfgName 
 		logMsg := utils.GetClockClassLogMessage(PTP4lProcessName, cfgName, clockClass)
 		e.writeLogToSocket(logMsg)
 	}
-	if !e.stdoutToSocket && e.clockClassMetric != nil {
+	if e.clockClassMetric != nil {
 		e.clockClassMetric.With(prometheus.Labels{
 			"process": PTP4lProcessName, "config": cfgName, "node": e.nodeName}).Set(float64(clockClass))
 	}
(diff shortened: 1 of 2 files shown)


aneeshkp:
> Root cause: The else-if / if !stdoutToSocket condition treats socket writes and Prometheus metric updates as mutually exclusive, but they serve different consumers and should both be active

 
I don’t think this fix is correct as written. When stdoutToSocket=true, cloud-event-proxy owns the Prometheus metrics context, not linuxptp-daemon. So updating clockClassMetric inside linuxptp-daemon will not update the metric that Prometheus actually scrapes. The fix should ensure the clock class value is sent through the socket/event path and then updated by cloud-event-proxy, which owns the Prometheus endpoint in this mode.

aabughosh:
Closing this PR — the `!stdoutToSocket` guard is intentional. When cloud-event-proxy is deployed, the daemon doesn't serve the metrics endpoint (port 9091) or register Prometheus metrics at all. cloud-event-proxy owns the Prometheus endpoint in that mode. The correct fix path is ensuring clock class values are correctly sent through the socket to cloud-event-proxy, which is addressed by PR #184.

PR #204, opened 2026-05-10
OCPBUGS-85361: Fix ptp and hardware profile matching logic

vitus133:
Make all instances of profile matching logic use ProfileNamesMatch function. That fixes mismatches when profile names are prefixed with resource name.

Assisted by Claude

### DIFF
--- a/pkg/controller/hardwareconfig_controller.go
+++ b/pkg/controller/hardwareconfig_controller.go
@@ -23,6 +23,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/daemon"
+	hc "github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/hardwareconfig"
 	ptpv1 "github.com/k8snetworkplumbingwg/ptp-operator/api/v1"
 	ptpv2alpha1 "github.com/k8snetworkplumbingwg/ptp-operator/api/v2alpha1"
 )
@@ -408,14 +409,8 @@ func (r *HardwareConfigReconciler) recordUpdateFailure(ctx context.Context, conf
 
 // ptpProfileNameMatchesActive reports whether relatedProfile is considered active.
 func ptpProfileNameMatchesActive(relatedProfile string, activePTPProfiles map[string]bool) bool {
-	if activePTPProfiles[relatedProfile] {
-		return true
-	}
-	// Also accept a match when an active profile entry carries the ptpconfig resource name
-	// as a prefix separated by "_".
-	suffix := "_" + relatedProfile
 	for activeProfile := range activePTPProfiles {
-		if strings.HasSuffix(activeProfile, suffix) {
+		if hc.ProfileNamesMatch(activeProfile, relatedProfile) {
 			return true
 		}
 	}
--- a/pkg/hardwareconfig/clockchain_resolution.go
+++ b/pkg/hardwareconfig/clockchain_resolution.go
@@ -157,6 +157,19 @@ func findLeadingInterfaceFromUpstreamPortWithResolver(upstreamPort string, resol
 	return leadingInterface, nil
 }
 
+// ProfileNamesMatch reports whether a PtpConfig profile name (stored) matches a
+// HardwareConfig's RelatedPtpProfileName (requested).
+func ProfileNamesMatch(stored, requested string) bool {
+	if strings.TrimSpace(requested) == "" {
+		return false
+	}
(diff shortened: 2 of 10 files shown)


pkg/controller/hardwareconfig_controller.go
```diff
@@ -408,14 +409,8 @@ func (r *HardwareConfigReconciler) recordUpdateFailure(ctx context.Context, conf
...
-	}
-	// Also accept a match when an active profile entry carries the ptpconfig resource name
-	// as a prefix separated by "_".
-	suffix := "_" + relatedProfile
 	for activeProfile := range activePTPProfiles {
-		if strings.HasSuffix(activeProfile, suffix) {
+		if hc.ProfileNamesMatch(activeProfile, relatedProfile) {
 			return true
 		}
 	}
```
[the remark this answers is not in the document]
vitus133: ProfileNamesMatch is also O1:
```
func ProfileNamesMatch(stored, requested string) bool {
       return stored == requested || strings.HasSuffix(stored, "_"+requested)
}
```

pkg/hardwareconfig/clockchain_resolution.go
```diff
@@ -157,6 +157,12 @@ func findLeadingInterfaceFromUpstreamPortWithResolver(upstreamPort string, resol
 	return leadingInterface, nil
 }
 
+// ProfileNamesMatch reports whether a PtpConfig profile name (stored) matches a
+// HardwareConfig's RelatedPtpProfileName (requested).
+func ProfileNamesMatch(stored, requested string) bool {
+	return stored == requested || strings.HasSuffix(stored, "_"+requested)
```
gtannous-spec: I guess there is a false-positive risk here, when the string `requested` itself contains underscores, a shorter substring can match falsely (unless profile names reject underscores from the first place)

pkg/hardwareconfig/clockchain_resolution.go
```diff
@@ -157,6 +157,12 @@ func findLeadingInterfaceFromUpstreamPortWithResolver(upstreamPort string, resol
 	return leadingInterface, nil
 }
 
+// ProfileNamesMatch reports whether a PtpConfig profile name (stored) matches a
+// HardwareConfig's RelatedPtpProfileName (requested).
+func ProfileNamesMatch(stored, requested string) bool {
```
gtannous-spec, line 162: shouldn't we consider adding here a guard checking whether `requested` is an empty string and return false in case ?

pkg/hardwareconfig/clockchain_resolution.go
vitus133: Hm. I was not following fully your discussions with @nocturnalastro on constraints of profile name concatenation. What was the outcome? Should we validate the profile or the resource name on admission to disallow underscores? Because if we don't - there will be always ambiguity.  
Or should we document that `relatedPtpProfileName` contains concatenated "resource-name"_"profile-name"?

nocturnalastro: As far as I'm aware there is nothing stopping a profile name from having underscores as its just a regular string field unlike with CRD name. 

So might want to use something along the lines of `strings.SplitN(stored, 2)[1] == requested`

gtannous-spec: Profile name can contain underscores, but the CRs reject underscores at k8s-level.

the suffix could be a bit tricky in case you have in the same `ptpconfig` two profile names one which is a suffix of other for example:
ptpconfig named Y and two profile names bc_profile and profile. 
Y_profile and Y_bc_profile would be qualified to the ConfigMap, the function `ProfileNamesMatch` gets: `ProfileNamesMatch(Y,profile)` could end up having two qualified profile names with `_profile` suffix.

I would suggest splitting the string to two parts by the first occurrence of the first `"_"` and check if the second part is THE SUFFIX, simply following the fact the the CR name is always the prefix and it doesn't contain underscores at all

vitus133: I think the simplest and the most straightforward way would be to document that relatedPtpProfileName of HardwarfeConfig resource should contain `"ptpConfig resource Name"_"PTP profile name"`
WDYT? @nocturnalastro @gtannous-spec

vitus133: Or use the `strings.SplitN(stored, "_", 2)[1] == requested`
WDYT?

vitus133: Fixed

gtannous-spec: lgtm

nocturnalastro (approved):
/lgtm

PR #206, opened 2026-05-11
OCPBUGS-85377: Sanitize GNSS UART before starting gpsd

vitus133:
On platforms where the GNSS module is connected via UART, killing gpsd during configuration transitions can leave the interface in an inconsistent state. This prevents gpsd from reconnecting properly on the next start.

This commit adds an stty sane invocation prior to starting gpsd. This ensures the UART is in a clean operational state. No regressions were observed on USB-connected GNSS modules.

Assisted by Cursor


nocturnalastro (approved):
/lgtm

PR #207, opened 2026-05-13
CNF-20903: First check the SOURCE_GIT_COMMIT before using the embedded git hash

nocturnalastro:
This allows for better traceablilty when the build system copies the source code into a local git store before building.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -86,8 +86,11 @@ func (cp *cliParams) debugPrint() {
 }
 
 func main() {
-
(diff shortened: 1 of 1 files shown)


dpopsuev:
@nocturnalastro Thanks Micky! does the 'SOURCE_GIT_COMMIT' come into action during build time?

nocturnalastro:
@dpopsuev Yes it was on the images that I grabbed for 4.18-4.21 matched the label that art said they use to track the original commit

vitus133 (approved):

PR #209, opened 2026-05-17
CNF-23072: Allow linking interfaces by PHC, add HPE vendor defaults

vitus133:
Previously, interfaces were logically linked by a common name prefix. For
example, 'eno4' and 'eno5' were aliased as 'enox' and linked to the same
PTP hardware clock (PHC) island. This link is required in Telecom Boundary
Clock (T-BC) implementations to relate the upstream port to the leading
NIC port that exposes the PTP device.
    
However, some hardware implementations (such as the HPE ProLiant EL140 Gen12)
use different prefixes for ports belonging to the same NIC,
causing name-prefix association to fail.
    
Fix this by allowing the leading NIC's default interface and the upstream
port to be associated directly by their PHC.

### DIFF
--- a/pkg/alias/alias.go
+++ b/pkg/alias/alias.go
@@ -32,6 +32,13 @@ func GetAllAliases() map[string]string {
 	return storeInstance.getAllAliases()
 }
 
+// GetPhcGroup returns the PHC device path for the given interface, or empty
+// string if the interface was not registered. Two interfaces with the same
+// non-empty return value share the same PHC hardware.
+func GetPhcGroup(ifname string) string {
+	return storeInstance.getPhcGroup(ifname)
+}
+
 // Debug ...
 func Debug(logF func(string, ...any)) {
 	for ifName, alias := range storeInstance.aliases {
--- a/pkg/alias/manager.go
+++ b/pkg/alias/manager.go
@@ -42,7 +42,7 @@ func (m *store) calculateAliases() {
 
 	errs := make([]error, 0)
 
-	// storeInstance.aliases = make(map[string]string)
+	m.aliases = make(map[string]string)
 
 	seenAliases := make(map[string][]string)
 	for phcID, group := range m.interfacesByPhc {
--- a/pkg/alias/manager.go
+++ b/pkg/alias/manager.go
@@ -13,6 +13,7 @@ import (
 type store struct {
 	sync.RWMutex
 	interfacesByPhc map[string][]string
+	phcByIface      map[string]string
 	aliases         map[string]string
 }
 
@@ -20,6 +21,13 @@ func (m *store) addInterface(phc string, ifName string) {
 	m.Lock()
 	defer m.Unlock()
 	m.interfacesByPhc[phc] = append(m.interfacesByPhc[phc], ifName)
+	m.phcByIface[ifName] = phc
+}
+
+func (m *store) getPhcGroup(ifName string) string {
+	m.RLock()
+	defer m.RUnlock()
+	return m.phcByIface[ifName]
 }
 
 // calculateAliases calculates and sets the aliases.
@@ -113,6 +121,7 @@ func (m *store) clear() {
 	defer m.Unlock()
 	m.aliases = make(map[string]string)
 	m.interfacesByPhc = make(map[string][]string)
+	m.phcByIface = make(map[string]string)
 }
 
 func (m *store) getAllAliases() map[string]string {
@@ -125,4 +134,8 @@ func (m *store) getAllAliases() map[string]string {
 	return result
 }
 
-var storeInstance = store{aliases: make(map[string]string), interfacesByPhc: make(map[string][]string)}
+var storeInstance = store{
+	aliases:         make(map[string]string),
+	interfacesByPhc: make(map[string][]string),
+	phcByIface:      make(map[string]string),
+}
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -441,6 +441,16 @@ func (dn *Daemon) getInterfacesFromHardwareConfig(nodeProfile *ptpv1.PtpProfile)
 			// Get PHC ID for the interface
 			phcID := ptpnetwork.GetPhcId(networkInterface)
 
+			// Register in the alias store so convergeConfig can match this
+			// interface against ptp4l interfaces that share the same PHC (e.g.
+			// eno1 vs eth3 on an 8-port NIC where naming prefixes differ).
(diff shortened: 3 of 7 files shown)


pkg/daemon/daemon.go
```diff
@@ -441,6 +441,16 @@ func (dn *Daemon) getInterfacesFromHardwareConfig(nodeProfile *ptpv1.PtpProfile)
...
 
+			// Register in the alias store so convergeConfig can match this
+			// interface against ptp4l interfaces that share the same PHC (e.g.
+			// eno1 vs eth3 on an 8-port NIC where naming prefixes differ).
+			if phcID != "" {
+				alias.AddInterface(phcID, networkInterface)
+				glog.Infof("getInterfacesFromHardwareConfig: registered iface %s phc %s in alias store", networkInterface, phcID)
+			} else {
+				glog.Warningf("getInterfacesFromHardwareConfig: could not get PHC ID for iface %s, convergeConfig PHC fallback will not work", networkInterface)
+			}
```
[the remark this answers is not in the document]
vitus133: The ptp4l interface is already registered — RenderPtp4lConf() calls alias.AddInterface(iface.PhcId, iface.Name) for every interface section it parses (config.go:463), and RenderPtp4lConf() is called for every process config at daemon.go:1035 during applyNodePTPProfiles(). So phcByIface will be populated before any events arrive.
Added interface cleanup in applyNodePTPProfiles

pkg/alias/manager.go
```diff
@@ -13,13 +13,21 @@ import (
...
+	phcByIface      map[string]string
 	aliases         map[string]string
 }
 
 func (m *store) addInterface(phc string, ifName string) {
 	m.Lock()
 	defer m.Unlock()
 	m.interfacesByPhc[phc] = append(m.interfacesByPhc[phc], ifName)
+	m.phcByIface[ifName] = phc
+}
```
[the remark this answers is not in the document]
vitus133: The store is cleared on profile application, so it's not realistic

pkg/event/event_tbc.go
```diff
@@ -639,15 +639,24 @@ func (e *EventHandler) getLeadingInterfaceBC() string {
 func (e *EventHandler) convergeConfig(event EventChannel) EventChannel {
 	if event.ProcessName == PTP4lProcessName {
 		iface := event.IFace
+		ifacePhc := alias.GetPhcGroup(iface)
+		glog.Infof("convergeConfig: ptp4l iface=%s phcGroup=%q original cfgName=%s", iface, ifacePhc, event.CfgName)
 		for cfg, dd := range e.data {
```
gtannous-spec, line 644: Do we consider the limitation that the system guarantees one DPLL config per PHC device? if not this can cause to non-deterministic results across runs. 
because `even.CfgName` is eventually overwritten, so if multiple configs have DPLL entries matching the same ptp4l interface, the last config iterated over a map silently wins since iterating through a map is non-deterministic across runs.

pkg/hardwareconfig/hardware-vendor/hpe/EL140-Gen12/behavior-profiles.yaml
```diff
@@ -0,0 +1,96 @@
+# Behavior profiles for Intel E825 hardware
```
gtannous-spec, line 1: shouldn't this be:
`# Behavior profiles for HPE EL140-Gen12 hardware`

pkg/hardwareconfig/hardware-vendor/hpe/EL140-Gen12/delays.yaml
```diff
@@ -0,0 +1,95 @@
+# Delay Compensation Model for Intel E825 (GNR-D)
```
gtannous-spec, line 1: shouldn't this be:
`# Delay Compensation Model for HPE EL140-Gen12`

pkg/alias/alias.go
```diff
@@ -32,6 +32,13 @@ func GetAllAliases() map[string]string {
 	return storeInstance.getAllAliases()
 }
 
+// GetPhcGroup returns the PHC device path for the given interface, or empty
+// string if the interface was not registered. Two interfaces with the same
+// non-empty return value share the same PHC hardware.
+func GetPhcGroup(ifname string) string {
```
gtannous-spec, line 38: Maybe we should consider adding test coverage for this API or PHC matching path, WDYT?

pkg/event/event_tbc.go
```diff
@@ -639,15 +639,24 @@ func (e *EventHandler) getLeadingInterfaceBC() string {
...
 			for _, item := range dd {
 				if item.ProcessName != DPLL {
 					continue
 				}
 				for _, dp := range item.Details {
-					if alias.GetAlias(dp.IFace) == alias.GetAlias(iface) {
+					dpPhc := alias.GetPhcGroup(dp.IFace)
+					aliasMatch := alias.GetAlias(dp.IFace) == alias.GetAlias(iface)
```
gtannous-spec, line 651: `GetAlias(dp.Iface)` and `GetAlias(iface)` are called more than one time, could probably be more effecient storing them in variable once and then reading them..

nocturnalastro (approved):
/lgtm

pkg/event/event_tbc.go
```diff
@@ -639,15 +639,24 @@ func (e *EventHandler) getLeadingInterfaceBC() string {
...
+					samePhc := dpPhc != "" && dpPhc == ifacePhc
+					glog.Infof("convergeConfig: checking DPLL iface=%s phcGroup=%q alias=%q vs ptp4l alias=%q aliasMatch=%v samePhc=%v",
```
gtannous-spec, line 653: We're doing excessive logging in the 3 iteration loops, the per-iteration log inside the DPLL detail loop fires on every ptp4l event which can lead to lots of unconditional log lines per event overhead.

If that would be redundant, maybe we should consider demoting inner-loop logs to glog(V2).Infof.
WDYT?

pkg/event/event_tbc.go
vitus133: This is not a realistic scenario

pkg/event/event_tbc.go
vitus133: dp.Iface is different every time in this loop

pkg/event/event_tbc.go
vitus133: log verbosity should be implemented first on the project level to utilize glog(V2)

PR #210, opened 2026-05-20
fix: display original profile names in config headers and logs

gtannous-spec:
### Summary
1. **ptp4l/synce4l** config files now render the #profile: header with the original profile name instead of the internal qualified form (ptpconfig_profileName).
2. **Daemon logs** across profile apply, HA failover, and hardwareconfig controller now display human-readable profile names — stripping the internal ptpconfig_ prefix so they match what the user originally configured.

### DIFF
--- a/pkg/controller/hardwareconfig_controller.go
+++ b/pkg/controller/hardwareconfig_controller.go
@@ -407,6 +407,21 @@ func (r *HardwareConfigReconciler) recordUpdateFailure(ctx context.Context, conf
 	// For now, we rely on events for error reporting
 }
 
+// formatQualifiedProfiles formats qualified profile names (ptpconfig_profileName)
+// into human-readable strings: "profile 'X' (ptpconfig 'Y'), ..."
+func formatQualifiedProfiles(qualifiedNames []string) string {
+	parts := make([]string, 0, len(qualifiedNames))
+	for _, qn := range qualifiedNames {
+		crName, profileName := hc.SplitQualifiedName(qn)
+		if crName != "" {
+			parts = append(parts, fmt.Sprintf("profile '%s' (ptpconfig '%s')", profileName, crName))
+		} else {
+			parts = append(parts, fmt.Sprintf("profile '%s'", qn))
+		}
+	}
+	return "[" + strings.Join(parts, ", ") + "]"
+}
+
 // ptpProfileNameMatchesActive reports whether relatedProfile is considered active.
 func ptpProfileNameMatchesActive(relatedProfile string, activePTPProfiles map[string]bool) bool {
 	for activeProfile := range activePTPProfiles {
@@ -432,7 +447,7 @@ func (r *HardwareConfigReconciler) calculateNodeHardwareConfigs(_ context.Contex
 		for _, profile := range activeProfiles {
 			activePTPProfiles[profile] = true
 		}
-		glog.Infof("Filtering hardware configs by active PTP profiles: %v", activeProfiles)
+		glog.Infof("Filtering hardware configs by active PTP profiles: %s", formatQualifiedProfiles(activeProfiles))
 	} else {
 		// No active profiles - exclude all hardware configs
 		activePTPProfiles = make(map[string]bool) // Empty map to trigger filtering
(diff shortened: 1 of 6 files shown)


pkg/controller/hardwareconfig_controller.go
```diff
@@ -407,6 +407,20 @@ func (r *HardwareConfigReconciler) recordUpdateFailure(ctx context.Context, conf
 	// For now, we rely on events for error reporting
 }
 
+// formatQualifiedProfiles formats qualified profile names (ptpconfig_profileName)
+// into human-readable strings: "profile 'X' (ptpconfig 'Y'), ..."
+func formatQualifiedProfiles(qualifiedNames []string) string {
+	parts := make([]string, 0, len(qualifiedNames))
+	for _, qn := range qualifiedNames {
+		if sep := strings.SplitN(qn, "_", 2); len(sep) == 2 {
```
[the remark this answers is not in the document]
gtannous-spec: Standardized the logic for splitting and matching qualified profile names into a function`SplitQualifiedName` which could be called where we want to duplicate this logic.

pkg/daemon/config.go
```diff
@@ -434,7 +447,7 @@ func getSectionName(name string) string {
 
 // RenderPtp4lConf outputs ptp4l config as string
 func (conf *Ptp4lConf) RenderPtp4lConf() (configOut string, ifaces config.IFaces) {
-	configOut = fmt.Sprintf("#profile: %s\n", conf.profile_name)
+	configOut = fmt.Sprintf("#profile: %s\n", DisplayProfileName(conf.profile_name))
```
nocturnalastro: This will likely break tests, this is also not really for human consumption

nocturnalastro: Actually no it won't but break them but it would break my T-BC PR as I added in a function for adding the prefix

https://github.com/k8snetworkplumbingwg/ptp-operator/pull/205/changes/d4b53625581ce00d018b70900e1932de3bbc04b9#diff-e36981059d6c4063ab4c9b720250ec60668714cf4386489f5ed7e558b62b189bR1040

nocturnalastro: That being said the reason we wanted the prefix was to remove possible ambiguities we shouldn't add them back just to make it a little easier to read.

pkg/daemon/daemon.go
```diff
@@ -718,37 +718,38 @@ func (dn *Daemon) applyNodePTPProfiles() error {
...
 	// (needed to keep code compatibility elsewhere and allow it to work both with hardware config and plugins)
 	for _, profile := range dn.ptpUpdate.NodeProfiles {
-		glog.Infof("Processing profile: %s", *profile.Name)
+		displayName := DisplayProfileName(*profile.Name)
+		glog.Infof("Processing profile: %s", displayName)
 
 		// Log profile details for debugging
 		if profile.Interface != nil {
-			glog.Infof("Profile %s interface: %s", *profile.Name, *profile.Interface)
+			glog.Infof("Profile %s interface: %s", displayName, *profile.Interface)
```
nocturnalastro: Same as above I don't see how this is an improvement here as we could have two profiles with the same name in different configs.

gtannous-spec: in fact displayname called at `displayName := DisplayProfileName(*profile.Name)` , this function calls `SplitQualifiedName` which returns two variables (CR name , profile name) 
what if instead we store also the first returned variable and call it through the log with something like: 
Profile %s from ptpconfig %s %interface %s, so we log the profile name + which ptpconfig it was defined at instead of logging crName + "_" + profileName (as a whole string), that could easier to understand if someone who's checking the logs is not aware of the fact that we are logging the qualified profile name with the underscore separator? 

unless we doc it and explicitly say that the *profileNames are consisted of (CrName + _ + originalProfileName) 
WDYT? @nocturnalastro

pkg/daemon/config.go
gtannous-spec: Yep I agree on that one, thought it would just be easier (to read the logs) for others who are not aware of this change :) 
if it doesn't really matter then yeah we don't actually need this changes..

pkg/daemon/daemon.go
nocturnalastro: I think that makes sense just update the logs to take both :)

gtannous-spec: Does your test use regex to check if the logs contain certain qualified profile name?

gtannous-spec: Will also update config header files to start with : #Profile X from ptpconfig Y 
instead of #profile X_Y.
@nocturnalastro does it make sense?

gtannous-spec: Now logs contain both :) 
rebased and squashed to a single commit

gtannous-spec:
Would be simpler to write a doc explaining the "qualified profile names" rather than extracting out the original profile names.

PR #211, opened 2026-05-20
 Add in-tree implementation of cloud-event-proxy

sebsoto:
Replaces the out-of-tree cloud-event-proxy sidecar with a lightweight
implementation built directly into linuxptp-daemon. The daemon
communicates PTP state changes over a Unix socket using NDJSON-encoded
IPC messages, and the proxy exposes them as O-RAN WG6 CloudEvents via
a REST API.

This is not yet used, the existing cloud-event-proxy is still in use,
further changes to the daemon are required to make use of this new
implementation.


cmd/cloud-event-proxy/main.go
```diff
@@ -0,0 +1,93 @@
...
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+	go func() {
+		sig := <-sigCh
+		log.Infof("received signal %s, shutting down", sig)
+		close(closeCh)
+	}()
+
+	cache := cep.NewEventCache(nodeName)
+	ps := cep.NewPubSub(filepath.Join(storePath, "subscriptions.json"), func(endpoint string) io.Writer {
```
nocturnalastro: Looks this and the equivalent in the tests are the same, why shouldn't this be a common function something like `HTTPWriterFactory`?

pkg/cep/cep.go
```diff
@@ -0,0 +1,384 @@
...
+		ValueType: event.DECIMAL,
+		Value:     value,
+	}
+}
+
+// TODO: replace placeholder offsets with real values from the daemon via IPC.
+// Offsets are not part of the O-RAN spec but were included in cloud-event-proxy v1.
+// Including offsets in state change events either makes them noisier (firing on every
+// offset change) or loses offset tracking when the state is not changing. A separate
+// IPC message for periodic offset reporting would avoid both problems and should be considered.
```
nocturnalastro, line 79: +1

cmd/cloud-event-proxy/main.go
nocturnalastro:
```go
func NewHTTPWriterFactory(endpoint string, timeout time.Duration) func(string) io.Writer {
	return func(endpoint string) io.Writer {
		return NewHTTPWriter(endpoint, &http.Client{Timeout: timeout})
	}
}
```

pkg/cep/event_cache.go
```diff
@@ -0,0 +1,130 @@
...
+		e = &event.Event{
+			Type:            string(oranType),
+			Source:          path.Join("/cluster/node", c.nodeName, string(oranSource)),
+			DataContentType: &contentType,
+			Data:            &event.Data{Version: event.APISchemaVersion},
+		}
+		c.entries[msg.Type] = e
+	}
+
+	var changed bool
```
nocturnalastro, line 58: Why not put a diff or equal function which could be used define changes for its type.

cmd/cloud-event-proxy/main.go
```diff
@@ -0,0 +1,89 @@
...
+	"net"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"syscall"
+	"time"
+
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/cep"
+
+	log "github.com/sirupsen/logrus"
```
nocturnalastro: Perhaps we should use glog to make all the code consistent

jzding (approved):
/lgtm

PR #212, opened 2026-05-20
OCPBUGS-85092: Move sendSidecarRestart before process start

edcdavid:
## Summary

- Move `sendSidecarRestart()` from the end of `applyNodePTPProfiles()` to right after `stopAllProcesses()`, before config files are written and processes start.
- Add a 1-second sleep after sending CMD RESTART to allow the sidecar to exec and rebind its socket listener.
- This ensures the sidecar is ready to receive all port-state and clock-sync events from ptp4l's first log line.

## Root Cause

PR #155 (`70f8cd2c`) placed `sendSidecarRestart()` at the end of `applyNodePTPProfiles()`, after ptp4l is already running. This creates a race condition where ptp4l can emit port-state transitions (`FAULTY -> SLAVE`) while the sidecar is mid-restart. The sidecar misses these events, leaving `masterOffsetIface` empty and `openshift_ptp_clock_state` stuck at `FREERUN`.

This regresses the fix from PR #29 (OCPBUGS-54967) which correctly delays profile loading until the sidecar connection is ready.

## Fix

The fix reorders operations in `applyNodePTPProfiles()`:

**Before (broken):**
```
stopAllProcesses() → write configs → start ptp4l → sendSidecarRestart()
```

**After (fixed):**
```
stopAllProcesses() → sendSidecarRestart() → sleep(1s) → write configs → start ptp4l
```

The sidecar's `syscall.Exec` and socket rebind completes well within 1 second since the binary is already in memory.

## Test Plan

- [ ] Verify clock_state returns to LOCKED after PTP threshold modify+restore with 2-port OC configuration
- [ ] Verify no port-state events are lost during reconfiguration
- [ ] Verify sidecar correctly receives all ptp4l logs after restart

Fixes: https://issues.redhat.com/browse/OCPBUGS-85092


edcdavid:
Closing in favor of a rebased PR on upstream/main.

PR #216, opened 2026-05-20
OCPBUGS-85092: add live gate to ensure replay completes before live data

edcdavid:
## Root Cause

After a PTP reconfiguration, linuxptp-daemon sends `CMD RESTART` to cloud-event-proxy (CEP), which execs itself and re-requests historical state via `GET /emit-logs`. The `/emit-logs` handler writes replay data (clock sync state + port roles) through the EventHandler's socket connection, while ptp4l socket-writer goroutines simultaneously forward live data on their own connections. CEP accepts both connections independently with no ordering guarantee.

If a stale replayed port role transition (e.g. `SLAVE→FAULTY` from before recovery) is processed by CEP after live data, CEP transitions `openshift_ptp_clock_state` to HOLDOVER/FREERUN — because port roles gate the clock state machine: LOCKED requires SLAVE, and a stale FAULTY removes that prerequisite. Unlike offset lines (which arrive every second), port role events only fire on actual ptp4l state transitions, so there is no natural corrective event. The metric becomes permanently stuck.

## Fix

Add a channel-based live gate that blocks ptp4l socket-writer goroutines from writing to the event socket until `/emit-logs` finishes replaying historical state. This guarantees CEP processes replay before live data, so the live (correct) state is always the final state.

Key changes:

- **Live gate** (`resetLiveGate` / `openLiveGate` / `waitForLiveGate`): channel + sync.Once mechanism with a 60s safety timeout. Reset in `applyNodePTPProfiles` before processes start; opened by `/emit-logs` after synchronous replay completes.
- **Split `cmdRun` goroutine**: the per-process goroutine is now two — a scanner (drains ptp4l stdout immediately to prevent pipe stalls) and a socket-writer (blocks on gate, then forwards). Previously a single goroutine did both, which would deadlock if gated.
- **`CMD LIVE_START` marker**: sent by each socket-writer after the gate opens, giving CEP explicit visibility into the replay/live boundary.
- **Synchronous `/emit-logs`**: `EmitClockSyncLogs` + `EmitPortRoleLogs` run synchronously; the gate opens only after both complete. Non-critical emits (ProcessStatus, ClockClass) remain async.
- **`sendSidecarRestart` → Daemon method**: receives the Daemon receiver so it can access the gate.
- **`SetPortRole` mutex**: adds missing lock around the portRole map write.

## Unit & Integration Tests Added

- `TestLiveGate_NilGateReturnsImmediately` — calling waitForLiveGate before the gate is initialized is a safe no-op
- `TestLiveGate_OpenBeforeWait` — pre-opened gate returns immediately
- `TestLiveGate_WaitThenOpen` — blocked waiter unblocks when gate opens
- `TestLiveGate_ClosedGateBlocksUntilTimeout` — safety timeout fires when gate never opens
- `TestLiveGate_DoubleOpenIsSafe` — idempotent open via sync.Once
- `TestLiveGate_ResetReblocks` — reset after open creates a new blocking gate
- `TestLiveGate_MultipleWaitersAllUnblock` — all goroutines proceed on single open
- `TestLiveGateIntegration_FullDaemonCEPFlow` — end-to-end: ptp4l stdout → scanner → lineCh → socket-writer → CEP, with /emit-logs triggering gate open
- `TestLiveGateIntegration_ScannerDrainsWhileGateBlocked` — proves scanner never blocks on gate (no ptp4l pipe stall)
- `TestLiveGateIntegration_ReplayThenLive` — two-connection scenario with /emit-logs HTTP call, verifies both replay and live data arrive on separate connections
- `TestLiveGateIntegration_OrderingPreventsCorruption` — verifies the core guarantee: replay data is fully received before live data when the gate synchronizes write order
- `TestLiveGateIntegration_TriggerLogsFailureRecovery` — when /emit-logs fails (daemon not ready), socket-writer recovers on next connection
- `TestEmitLogsHandler_OpensGateAfterReplay` — /emit-logs handler opens gate synchronously after replay
- `TestLiveGate_TimeoutReturnsTrue` — verifies the safety timeout unblocks without hanging forever

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -131,12 +131,9 @@ func dialSocket() (net.Conn, error) {
 // over a short-lived dedicated connection to the event socket. The sidecar will exec itself
 // for a clean restart, then re-read all configuration from disk (ConfigMap + ptp4l config files).
 //
-// Before sending CMD RESTART, the live gate is reset so that ptp4l process connections
-// block until the next /emit-logs (replay) completes. This guarantees CEP sees replay
-// state before any live data.
+// The live gate is reset earlier in applyNodePTPProfiles (before processes start)
+// so that socket-writers block until the next /emit-logs (replay) completes.
 func (dn *Daemon) sendSidecarRestart() error {
-	dn.resetLiveGate()
-
 	c, err := net.Dial("unix", eventSocket)
 	if err != nil {
 		return err
@@ -426,27 +423,21 @@ func (dn *Daemon) openLiveGate() {
 // guarantee. Exposed as a var so tests can shorten it.
 var liveGateTimeout = 60 * time.Second
 
-// waitForLiveGate blocks until the gate opens, exitCh fires, or a safety
-// timeout expires. Returns true if gate opened, false if exit/timeout.
-func (dn *Daemon) waitForLiveGate(exitCh chan bool) bool {
+// waitForLiveGate blocks until the gate opens or a safety timeout expires.
+func (dn *Daemon) waitForLiveGate() {
 	dn.liveGateMu.Lock()
 	gate := dn.liveGate
 	dn.liveGateMu.Unlock()
 	if gate == nil {
 		glog.V(4).Info("waitForLiveGate: gate is nil, returning immediately")
-		return true
+		return
 	}
 	glog.V(4).Info("waitForLiveGate: blocking until gate opens")
 	select {
 	case <-gate:
 		glog.V(4).Info("waitForLiveGate: gate opened, proceeding")
-		return true
-	case <-exitCh:
-		glog.V(4).Info("waitForLiveGate: exitCh fired, aborting")
-		return false
 	case <-time.After(liveGateTimeout):
 		glog.Warning("liveGate: timeout after 60s, proceeding without replay guarantee")
-		return true
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/daemon/daemon.go
```diff
@@ -1597,51 +1666,103 @@ func (p *ptpProcess) cmdRun(stdoutToSocket bool, pm *plugin.PluginManager) {
...
+						p.announceHAFailOver(nil, output)
+					}
+					select {
+					case lineCh <- output:
+					default:
+						glog.Warning("liveGate: lineCh full, dropping line for socket forwarding")
+					}
+				}
+				close(lineCh)
+			}()
```
nocturnalastro: Could this be on the process itself?
This function is quite large and it can be hard to tell whats happening. 
Perhaps we should have a tech debt story for that?

pkg/daemon/daemon.go
```diff
@@ -1597,51 +1666,103 @@ func (p *ptpProcess) cmdRun(stdoutToSocket bool, pm *plugin.PluginManager) {
...
+
+			// Socket-writer goroutine: connects to the event socket, waits for
+			// the live gate (replay) to complete, then forwards lines to CEP.
```
nocturnalastro: What happens if linuxptp-daemon container crash's do we have a guarantee that cloud-event-proxy will request the log replay?

edcdavid: If linuxptp-daemon container crashes, it calls applyNodePTPProfiles which will start the sync from scratch (replay+livegate)

pkg/daemon/daemon.go
edcdavid: Fixed

nocturnalastro (approved):
/lgtm

PR #222, opened 2026-06-09
Raise replay-fix log verbosity above production threshold

edcdavid:
The live gate and socket-writer logs introduced by the replay-fix (PR #216)
use glog.V(4), which fires at the production pod setting of -v=10. This
creates excessive log noise in normal operation.

Raise all new replay-fix debug logs from V(4) to V(14) so they remain silent
at production verbosity but are still available for troubleshooting with
-v=14 or higher.

Pre-existing logs that were inadvertently downgraded during the refactoring
(socket write errors) are restored to their original Error level.


nocturnalastro (approved):
/lgtm

PR #223, opened 2026-06-14
OCPBUGS-88512: Fix gnr-d initialization

vitus133:
Before the fix, both T-GM and T-BC left hardware configurations that could interfere with the other profile.
- T-GM e825 plugin initialized all pins (including both PTP input pins) as selectable for both eec and pps parent devices.
- TBC enabled 1PPS and 1kHz outputs to DPLL

These configurations impacted the clock operation in the following ways:
- T-BC had EEC input enabled and locking, creating a parasitic feedback loop resulted in `ptp4l` offset oscillations and the servo failing to converge
- T-GM was working with PHC sync outputs available for DPLL to consider as a source. As a result, the eec could lock on GNSS and the pps - pn the PHC

The fix:
- T-GM will disable e825 outputs to DPLL on init
- T-BC eill disable both GNSS inputs and the eec input of both PTP pins

Assisted by Cursor

### DIFF
--- a/addons/intel/e825.go
+++ b/addons/intel/e825.go
@@ -62,6 +62,27 @@ func tbcConfigured(nodeProfile *ptpv1.PtpProfile) bool {
 	return nodeProfile.PtpSettings["clockType"] == "T-BC"
 }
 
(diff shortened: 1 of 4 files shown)


nocturnalastro (approved):
/lgtm

PR #224, opened 2026-06-18
Track active TR port during dual-upstream T-BC switchover for correct event reporting

gtannous-spec:
# Track active TR port on SLAVE selection for correct event/log reporting

## Summary

In T-BC deployments with two upstream (TR) ports, when the A-BMCA selects a backup port as SLAVE during switchover, the daemon continues reporting the original port in events and logs because `activePort` is never updated. This patch updates `activePort` on every `MASTER_CLOCK_SELECTED` transition so that `activeTRPort()` — used by both `sendPtp4lEvent()` and `sendPtp4lOffsetEvent()` — returns the correct active interface.
Jira Ticket: [CNF-25167](https://redhat.atlassian.net/browse/CNF-25167)

## Problem

When a T-BC profile has two TR-facing interfaces (dual-upstream for redundancy), the `processTBCTransitionLegacy` function never sets `activePort` when a port is selected as SLAVE. The `activeTRPort()` method falls back to `trIfaceNames[0]` (always the first configured port), so events and logs always report the main port — even after the system has switched over to the backup.

### `daemon.go`

When ptp4l logs `"to SLAVE on MASTER_CLOCK_SELECTED"`, extract the port name and update `activePort`. This ensures `activeTRPort()` returns the correct interface for all downstream event and log paths.

### `daemon_internal_test.go`

Four new tests covering dual-upstream switchover scenarios:

- **Switchover with activePort update** — active port goes FAULTY, enters holdover, backup becomes SLAVE, verifies `activePort` updates to the backup interface
- **All ports lost enters holdover** — active port loses SLAVE via announce timeout, verifies holdover entry
- **Recovery from holdover** — starting from holdover, backup becomes SLAVE, offset filter fills and converges, verifies holdover exit to LOCKED
- **Correct interface reporting** — verifies `activeTRPort()` returns the correct port after initial selection and after switchover

## Testing

- Verified on cnfdg42 with dual-upstream T-BC configuration (ens2f1, ens2f3)
- Confirmed events and logs report the correct active interface after switchover
- Confirmed holdover entry on port loss and recovery via offset filter convergence
- All existing and new unit tests pass

## Release Notes

* **Bug Fixes**
  * Improved legacy precision time protocol failover handling by more reliably detecting and updating the active upstream port from ptp-related output, including avoiding incorrect updates from non-relevant port name matches.

* **Tests**
  * Expanded legacy-mode test coverage for upstream switchover, failed switchover behavior, transition into holdover when all upstreams are lost, and the convergence requirements to recover from holdover, plus active port/interface reporting checks.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1589,6 +1589,11 @@ func (p *ptpProcess) processTBCTransitionLegacy(output string, pm *plugin.Plugin
 	}
 	if portMatched {
 		if strings.Contains(output, "to SLAVE on MASTER_CLOCK_SELECTED") {
+			portName := parser.ExtractPortName(output)
+			if portName != "" {
+				p.tBCAttributes.activePort = portName
+				glog.Infof("T-BC active upstream port: %s", portName)
+			}
 			p.tBCAttributes.lastReportedState = event.PTP_LOCKED
 			p.tBCAttributes.offsetFilter = utils.NewWindow(offsetFilterSize)
 		} else if strings.Contains(output, "to MASTER on ANNOUNCE_RECEIPT_TIMEOUT_EXPIRES") ||
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1591,8 +1591,12 @@ func (p *ptpProcess) processTBCTransitionLegacy(output string, pm *plugin.Plugin
 		if strings.Contains(output, "to SLAVE on MASTER_CLOCK_SELECTED") {
 			portName := parser.ExtractPortName(output)
 			if portName != "" {
-				p.tBCAttributes.activePort = portName
-				glog.Infof("T-BC active upstream port: %s", portName)
+				if len(p.tBCAttributes.trIfaceNames) > 1 && !slices.Contains(p.tBCAttributes.trIfaceNames, portName) {
+					glog.Warningf("Ignoring non-TR port in legacy MASTER_CLOCK_SELECTED event: %s", portName)
+				} else {
+					p.tBCAttributes.activePort = portName
+					glog.Infof("T-BC active upstream port: %s", portName)
+				}
 			}
 			p.tBCAttributes.lastReportedState = event.PTP_LOCKED
 			p.tBCAttributes.offsetFilter = utils.NewWindow(offsetFilterSize)
--- a/pkg/daemon/daemon_internal_test.go
+++ b/pkg/daemon/daemon_internal_test.go
@@ -1819,6 +1819,41 @@ func TestTBCLegacy_ActiveTRPort_ReportsCorrectInterface(t *testing.T) {
 		"activeTRPort should reflect the newly selected backup port")
 }
 
+func TestTBCLegacy_ActivePort_IgnoresNonTRPort(t *testing.T) {
+	pmStruct, _ := registerPlugins([]string{})
+	pm := &pmStruct
+
+	oldValue := vTbcHasHardwareConfig
+	vTbcHasHardwareConfig = false
+	defer func() { vTbcHasHardwareConfig = oldValue }()
+
+	process := &ptpProcess{
+		tBCAttributes: tBCProcessAttributes{
+			trIfaceNames:      []string{testDUTUpstream1, testDUTUpstream2},
+			perPortState:      map[string]event.PTPState{testDUTUpstream1: event.PTP_LOCKED, testDUTUpstream2: event.PTP_NOTSET},
+			activePort:        testDUTUpstream1,
+			trPortsConfigFile: "test-config",
+			lastReportedState: event.PTP_LOCKED,
+			lastAppliedState:  event.PTP_LOCKED,
+			offsetThreshold:   10.0,
+		},
+		nodeProfile: ptpv1.PtpProfile{
+			Name:        stringPointer("test-profile"),
+			PtpSettings: map[string]string{"leadingInterface": testDUTLeadingIface, testDUTClockIDKey: "123456789"},
+		},
+		eventCh:    make(chan event.Event, 10),
+		configName: "test-config",
+		clockType:  event.BC,
+	}
+
+	// A port with a prefix-colliding name (ens2f10 vs tracked ens2f1) fires MASTER_CLOCK_SELECTED.
+	// The log line contains "ens2f1" as a substring so portMatched is true, but
+	// ExtractPortName returns "ens2f10" which is NOT in trIfaceNames.
+	process.tBCTransitionCheck("ptp4l[300]: [test-config.0.config] port 3 (ens2f10): UNCALIBRATED to SLAVE on MASTER_CLOCK_SELECTED", pm)
+	assert.Equal(t, testDUTUpstream1, process.tBCAttributes.activePort,
+		"activePort must not change to a non-TR port with a prefix-colliding name")
+}
+
 // TestPtp4lConf_PopulatePtp4lConf_ClockTypeWithCliArgs tests clock_type detection with cliArgs parameter
 func TestPtp4lConf_PopulatePtp4lConf_ClockTypeWithCliArgs(t *testing.T) {
 	tests := []struct {
--- a/pkg/daemon/daemon_internal_test.go
+++ b/pkg/daemon/daemon_internal_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/event"
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/hardwareconfig"
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/leap"
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/utils"
 	ptpv1 "github.com/k8snetworkplumbingwg/ptp-operator/api/v1"
 	ptpv2alpha1 "github.com/k8snetworkplumbingwg/ptp-operator/api/v2alpha1"
 	"github.com/stretchr/testify/assert"
@@ -32,6 +33,13 @@ import (
 )
 
 const testPtp4lOffsetLine = "ptp4l[1.000]: [ptp4l.0.config] master offset 5 s2 freq -1000 path delay 100"
+
+const (
+	testDUTLeadingIface = "ens2f0"
+	testDUTUpstream1    = "ens2f1"
+	testDUTUpstream2    = "ens2f3"
+	testDUTClockIDKey   = "clockId[ens2f0]"
+)
 
 // vendor defaults are embedded; no filesystem setup needed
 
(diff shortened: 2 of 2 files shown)


pkg/daemon/daemon.go
```diff
@@ -1586,11 +1602,72 @@ func (p *ptpProcess) processTBCTransitionLegacy(output string, pm *plugin.Plugin
 		}
 	}
 	if portMatched {
+		if strings.Contains(output, "Master clock quality received is greater than configured, ignoring master") {
+			portName := parser.ExtractPortName(output)
+			if portName != "" && len(p.tBCAttributes.trIfaceNames) > 1 {
+				p.tBCAttributes.perPortState[portName] = event.PTP_FREERUN
```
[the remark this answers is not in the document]
gtannous-spec: If there are no upstream ports, neither `trIfaceNames` nor `perPortState` is initialized. Both paths that could write to the map are naturally guarded by the same condition,
Empty `trIfaceNames` means portMatched is always false, so the entire block that writes to perPortState is skipped, same for Hardwardconfig paths, without upstream ports, tbcStateDetector is nil, and the hardware config path is never taken.. 
There is no execution path where perPortState is written to without first being initialized. 
The existing guards are sufficient !!

pkg/daemon/daemon_internal_test.go
```diff
@@ -1298,6 +1306,217 @@ func TestTBCTransitionCheck_LegacyPath(t *testing.T) {
...
+	// testDUTUpstream2 still LOCKED
+	assert.Equal(t, event.PTP_LOCKED, process.tBCAttributes.perPortState[testDUTUpstream2])
+	// Aggregate state should NOT have dropped to FREERUN
+	assert.Equal(t, event.PTP_LOCKED, process.tBCAttributes.lastReportedState,
+		"should stay LOCKED when another TR port is still active")
+	// Should NOT have entered holdover
+	assert.NotEqual(t, event.PTP_HOLDOVER, process.tBCAttributes.lastAppliedState,
+		"should not enter holdover when backup port is still active")
+	// offsetFilter should NOT have been nulled
+	assert.Equal(t, event.PTP_LOCKED, process.tBCAttributes.lastAppliedState)
```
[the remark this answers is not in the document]
gtannous-spec: Fixed !

pkg/daemon/daemon_internal_test.go
```diff
@@ -1298,6 +1306,217 @@ func TestTBCTransitionCheck_LegacyPath(t *testing.T) {
...
+	// Fill offset filter to complete transition
+	for i := 0; i < 64; i++ {
+		process.tBCTransitionCheck("ptp4l[300]: [test-config.0.config] master offset 5 s2 freq 0 path delay 100", pm)
+	}
+	assert.Equal(t, event.PTP_LOCKED, process.tBCAttributes.lastAppliedState)
+
+	// Now test the REVERSE switchover: testDUTUpstream1 comes back up, A-BMCA switches back
+	process.tBCTransitionCheck("ptp4l[400]: [test-config.0.config] port 1 ("+testDUTUpstream1+"): FAULTY to LISTENING on INIT_COMPLETE", pm)
+	assert.Equal(t, event.PTP_NOTSET, process.tBCAttributes.perPortState[testDUTUpstream1],
+		"LISTENING should reset stale FREERUN to NOTSET")
```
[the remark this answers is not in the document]
gtannous-spec: Fixed !

pkg/daemon/daemon.go
```diff
@@ -1586,11 +1602,72 @@ func (p *ptpProcess) processTBCTransitionLegacy(output string, pm *plugin.Plugin
...
+					p.tBCAttributes.activePort = ""
+					pm.AfterRunPTPCommand(&p.nodeProfile, "tbc-ho-entry")
+					p.tBCAttributes.lastReportedState = event.PTP_FREERUN
+					glog.Info("T-BC no locked port remaining after quality rejection, MOVE TO HOLDOVER")
+					p.sendPtp4lEvent()
+					p.tBCAttributes.lastAppliedState = event.PTP_HOLDOVER
+					p.tBCAttributes.offsetFilter = nil
+				} else {
+					glog.Infof("T-BC port %s rejected but another TR port still available, staying in current state", portName)
+				}
```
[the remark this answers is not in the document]
gtannous-spec: Fixed.

pkg/daemon/daemon.go
```diff
@@ -1586,11 +1602,75 @@ func (p *ptpProcess) processTBCTransitionLegacy(output string, pm *plugin.Plugin
...
+		}
 		if strings.Contains(output, "to SLAVE on MASTER_CLOCK_SELECTED") {
+			portName := parser.ExtractPortName(output)
+			if portName != "" && len(p.tBCAttributes.trIfaceNames) > 1 {
+				p.tBCAttributes.perPortState[portName] = event.PTP_LOCKED
+				p.tBCAttributes.activePort = portName
+			}
 			p.tBCAttributes.lastReportedState = event.PTP_LOCKED
 			p.tBCAttributes.offsetFilter = utils.NewWindow(offsetFilterSize)
 		} else if strings.Contains(output, "to MASTER on ANNOUNCE_RECEIPT_TIMEOUT_EXPIRES") ||
```
nocturnalastro, line 1603: @vitus133 I don't know why we aren't using a parser event here Looks like it happens before we produce them should we make a tech debt?

pkg/daemon/daemon.go
```diff
@@ -1586,11 +1602,75 @@ func (p *ptpProcess) processTBCTransitionLegacy(output string, pm *plugin.Plugin
...
 			strings.Contains(output, "SLAVE to") {
+			if len(p.tBCAttributes.trIfaceNames) > 1 {
+				portName := parser.ExtractPortName(output)
+				if portName != "" {
+					if strings.Contains(output, "FAULT_DETECTED") {
+						p.tBCAttributes.perPortState[portName] = event.PTP_FREERUN
```
nocturnalastro: I'm not sure using a PTPState to describe a port is the correct thing. 
Why would we not use the normal port states? 

Edit: But looks like it was already like that so NVM.

pkg/daemon/daemon_internal_test.go
```diff
@@ -1298,6 +1307,219 @@ func TestTBCTransitionCheck_LegacyPath(t *testing.T) {
 	})
 }
 
+func TestTBCLegacy_DualTR_SinglePortLoss_NoFreerun(t *testing.T) {
```
nocturnalastro: Could these be a table? Seems like a lot of repetition?

pkg/daemon/daemon_internal_test.go
```diff
@@ -1579,6 +1828,173 @@ func TestTBCDualUpstream_ActiveTRPort_Helper(t *testing.T) {
...
+			perPortState:      map[string]event.PTPState{"ens2f1": event.PTP_LOCKED},
+			activePort:        "ens2f1",
+			trPortsConfigFile: "test-config",
+			lastReportedState: event.PTP_LOCKED,
+			lastAppliedState:  event.PTP_LOCKED,
+			offsetThreshold:   10.0,
+		},
+		nodeProfile: ptpv1.PtpProfile{
+			Name:        stringPointer("test-profile"),
+			PtpSettings: map[string]string{"leadingInterface": "ens2f0", "clockId[ens2f0]": "123456789"},
```
nocturnalastro: testDUTUpstream0?

pkg/daemon/daemon_internal_test.go
```diff
@@ -1579,6 +1828,173 @@ func TestTBCDualUpstream_ActiveTRPort_Helper(t *testing.T) {
...
+		},
+		eventCh:    make(chan event.EventChannel, 10),
+		configName: "test-config",
+		clockType:  event.BC,
+		offset:     0.0,
+	}
+
+	// Single TR port rejects master quality — enter holdover immediately
+	process.tBCTransitionCheck("ptp4l[100]: [test-config.0.config:3] port 1 (ens2f1): Master clock quality received is greater than configured, ignoring master!", pm)
```
nocturnalastro: testDUTUpstream1?

pkg/daemon/daemon_internal_test.go
```diff
@@ -1298,6 +1307,219 @@ func TestTBCTransitionCheck_LegacyPath(t *testing.T) {
...
+		eventCh:    make(chan event.EventChannel, 10),
+		configName: "test-config",
+		clockType:  event.BC,
+		offset:     5.0,
+	}
+
+	// testDUTUpstream1 loses SLAVE — testDUTUpstream2 is not LOCKED, but we simulate it being LOCKED first
+	process.tBCAttributes.perPortState[testDUTUpstream2] = event.PTP_LOCKED
+
+	process.tBCTransitionCheck("ptp4l[100]: [test-config.0.config] port 1 ("+testDUTUpstream1+"): SLAVE to MASTER on ANNOUNCE_RECEIPT_TIMEOUT_EXPIRES", pm)
```
nocturnalastro: This is used a fair bit in the tests maybe the pattern could be a const and just us sprintf?

nocturnalastro (commented):
Some minor things.

pkg/daemon/daemon_internal_test.go
gtannous-spec: Sorry, could you give more context WDYM? 
Function name probably :) ?

pkg/daemon/daemon_internal_test.go
gtannous-spec: same as above

pkg/daemon/daemon_internal_test.go
gtannous-spec: Yep makes sense, Will refactor the UT

pkg/daemon/daemon_internal_test.go
gtannous-spec: Yes, will fix it.

gtannous-spec: Addressed some of the comments, PTAL @nocturnalastro

pkg/daemon/daemon_internal_test.go
```diff
@@ -1298,6 +1312,248 @@ func TestTBCTransitionCheck_LegacyPath(t *testing.T) {
...
+			name:       "dual-TR: one port rejects quality, other available — no HOLDOVER",
+			trIfaces:   []string{testDUTUpstream1, testDUTUpstream2},
+			perPort:    map[string]event.PTPState{testDUTUpstream1: event.PTP_LOCKED, testDUTUpstream2: event.PTP_NOTSET},
+			activePort: testDUTUpstream1,
+			logLines: []string{
+				tLog(100, 1, testDUTUpstream1, "Master clock quality received is greater than configured, ignoring master!"),
+			},
+			wantPerPort:      map[string]event.PTPState{testDUTUpstream1: event.PTP_FREERUN},
+			wantLastReported: event.PTP_LOCKED,
+			wantLastApplied:  event.PTP_LOCKED,
```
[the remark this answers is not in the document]
gtannous-spec: Fixed.

pkg/daemon/daemon_internal_test.go
```diff
@@ -1298,6 +1312,248 @@ func TestTBCTransitionCheck_LegacyPath(t *testing.T) {
...
+		false,
+	)
+	process.offset = 5.0
+
+	// testDUTUpstream1 loses SLAVE via BMCA — testDUTUpstream2 is NOTSET (available) → stay in current state
+	process.tBCTransitionCheck(tLog(100, 1, testDUTUpstream1, "SLAVE to MASTER on ANNOUNCE_RECEIPT_TIMEOUT_EXPIRES"), pm)
+	assert.Equal(t, event.PTP_LOCKED, process.tBCAttributes.lastReportedState)
+	assert.Equal(t, event.PTP_NOTSET, process.tBCAttributes.perPortState[testDUTUpstream1],
+		"BMCA-driven loss should set NOTSET, not FREERUN")
+
```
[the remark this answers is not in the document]
gtannous-spec: fixed.

pkg/daemon/daemon.go
sebsoto: +1 on this.

~Seems this should just be a bool? Either the port is working or its not.~

Oops, nope thats not the case. An enum would be really helpful here to understand the behavior.

gtannous-spec: Why though? the event states are being explicitly mentioned there..
I do actually agree that a map literals for TR port states would be more clear for the logs if that's the case.. 
WDYT do we actually need them? or the event states are clear enough @sebsoto

gtannous-spec: Actually I have updated the PR to let the log point to the active TR port which is selected by the A-BMCA to be `MASTER_CLOCK_SELECTED` , this parsing part is not relevant anymore.

gtannous-spec:
@coderabbitai Please redo the PR description

gtannous-spec:
@coderabbitai review

pkg/daemon/daemon.go
vitus133: Yep, probably

pkg/daemon/daemon.go
vitus133, line 1540: @gtannous-spec could it be possible to verify this path ( hardwareconfig path) as well? although it looks like it is already handled

vitus133 (commented):
I commented on the hardwareconfig path - could you please verify it works as intended?

pkg/daemon/daemon.go
gtannous-spec: The hwconfig path already has full test coverage for the switchover scenarios..
in those functions: `PortALost_PortBTakesOver ` , `BothPortsLost ` , `RecoveryAfterBothLost` at `daemon_internal_test.go`

vitus133: That's true, but I think a lab test is needed to declare a victory here

gtannous-spec: Are we talking about a T-BC with legacy plugins or T-BC with a hardwareconfig CRD ? 
because for legacy plugins it's already done.

gtannous-spec: Done ! verified with hwconfig path too

pkg/daemon/daemon.go
```diff
@@ -1589,6 +1589,11 @@ func (p *ptpProcess) processTBCTransitionLegacy(output string, pm *plugin.Plugin
 	}
 	if portMatched {
 		if strings.Contains(output, "to SLAVE on MASTER_CLOCK_SELECTED") {
+			portName := parser.ExtractPortName(output)
+			if portName != "" {
+				p.tBCAttributes.activePort = portName
+				glog.Infof("T-BC active upstream port: %s", portName)
+			}
```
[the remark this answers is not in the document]
gtannous-spec: Fixed !

pkg/daemon/daemon.go
vitus133: @gtannous-spec could you please open a tech debt story for that

vitus133 (dismissed):
LGTM, please open a tech debt story for replacing the string parsing by generic parser event in the legacy path

pkg/daemon/daemon.go
gtannous-spec: Yes, will open soon

nocturnalastro (approved):
/lgtm

PR #227, opened 2026-06-23
Fix gnss_status always -1 due to double-newline in poll line join

lack:
UbloxPollPushThread reads ubxtool output with `ReadString('\\n')`, so
each line stored in the buffer includes its trailing newline character.
`MonitorGNSSEventsWithUblox` then joined those lines with `"\n"`, doubling
every newline:

```
"UBX-NAV-STATUS:\n" + "\n" + "  iTOW ... gpsFix 3 ...\n" + "\n" + ...
```

`processGNSSLines` used a `bufio.Scanner` which strips newlines, so when
the scanner detected `"UBX-NAV-STATUS:"` and did a one-line lookahead, it
landed on the spurious empty line instead of the data line containing
`"gpsFix"`. `ExtractNavStatus("")` found no `"gpsFix"` field and always
returned -1, causing `sourceLost=true` and `gnss_status=-1` regardless of
the actual GPS fix state.

The bug was introduced in 9ba63265 ("Decouple GNSS and DPLL") when the
direct per-line `UbloxPollPull()` loop was refactored into the batch
`processGNSSLines()` approach.

Fix by removing the join/split roundtrip entirely: pass the `[]string`
slice directly to `processGNSSLines` instead of joining into a string and
rescanning. This also makes `ExtractNavStatus` and `ExtractOffset` simpler —
they now take a single line instead of a multi-line string with an internal
split that was always a no-op in practice. `ExtractLeapSec` already used
`[]string` and is now consistent with the others.

`processGNSSLines` now uses index-based slice access for lookahead instead
of a scanner, and calls `strings.TrimRight` to strip any trailing newline
from lines as produced by `UbloxPollPull`. The regression is covered by the
"locked with good fix" test case, which passes lines with trailing newlines
and asserts correct parsing.

## Summary by CodeRabbit

* **Refactor**
  * Updated GNSS message parsing to use structured line input for more efficient clock/status extraction.
  * Improved NAV status and offset extraction to read directly from a single relevant line, with stricter handling for missing or malformed values.
* **Tests**
  * Expanded unit test coverage for NAV status and clock offset parsing, including multiple error scenarios.
  * Updated GNSS processing tests to reflect the new line-based input format.

### DIFF
--- a/pkg/daemon/gpsd.go
+++ b/pkg/daemon/gpsd.go
@@ -1,11 +1,9 @@
 package daemon
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"fmt"
-	"io"
 	"net"
 	"os"
 	"os/exec"
@@ -257,7 +255,7 @@ func (g *GPSD) MonitorGNSSEventsWithUblox() {
 					lines = append(lines, line)
 				}
 				if len(lines) > 0 {
-					g.processGNSSLines(strings.NewReader(strings.Join(lines, "\n")))
+					g.processGNSSLines(lines)
 				}
 			case <-g.monitorCtx.Done():
 				doneFn()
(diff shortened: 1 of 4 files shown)


pkg/daemon/gpsd.go
```diff
@@ -267,35 +265,35 @@ func (g *GPSD) MonitorGNSSEventsWithUblox() {
...
+			if i+1 < len(lines) {
+				nOffset = ublox.ExtractOffset(strings.TrimRight(lines[i+1], "\n"))
 			}
 		} else if strings.Contains(line, "UBX-NAV-STATUS") {
-			if scanner.Scan() {
-				nStatus = ublox.ExtractNavStatus(scanner.Text())
+			if i+1 < len(lines) {
+				nStatus = ublox.ExtractNavStatus(strings.TrimRight(lines[i+1], "\n"))
 			}
 		} else if strings.Contains(line, "UBX-NAV-TIMELS") {
```
sebsoto, line 286: Think we might want to add a `UBX-NAV-TIMELS` message to the TestProcessGNSSLines unit test.

sebsoto (commented):
Thanks for fixing this!

sebsoto (approved):

PR #228, opened 2026-06-23
Fix empty iface name for GNSS and GM metrics

lack:
GetLeadingInterface() in pkg/config/config.go only matched event.GNSS
and event.PTP4l sources. For a GM ts2phc configuration, slave interfaces
(e.g. [eno8703], [enp108s0f0] with ts2phc.master 0) are assigned
event.PPS by getSource() in RenderPtp4lConf(). The GNSS master is the
[nmea] section which RenderPtp4lConf explicitly excludes from the ifaces
list (NmeaSectionName guard). As a result, GetLeadingInterface() never
found a matching interface and returned Iface{Name:""}.

This caused gmInterface="" in the GPSD struct, which propagated into
every GNSS event as IFace:"". The event handler's getLeadingInterface()
then found the GNSS data entry with Details[0].IFace="" and returned ""
as the leading interface. Since "" != LEADING_INTERFACE_UNKNOWN, the
guard at line 974 of event.go did not fire, and UpdateClockStateMetrics
was called with alias.GetAlias("")="" for both the gnss and GM process
labels, triggering the sanity check warning every second, permanently.

Fix by adding event.PPS to the GetLeadingInterface() match condition.
A ts2phc PPS slave interface is the leading interface for a GM clock --
it is the interface being synchronised from the GNSS/PPS signal -- so
it is the correct answer here.

Add pkg/config/config_test.go with TestGetLeadingInterface covering
GNSS, PTP4l, PPS (the previously broken case), mixed sources, and
empty/unmatched lists.

Assisted-by: Claude Sonnet 4.6 and pi.dev
Signed-off-by: Jim Ramsay <jramsay@redhat.com>

## Summary by CodeRabbit

* **New Features**
  * Expanded leading clock source detection to include Pulse Per Second (PPS) alongside existing Global Navigation Satellite System (GNSS) and Precision Time Protocol (PTP4l) support.

* **Tests**
  * Added comprehensive test coverage for clock source detection, validating multiple scenarios and edge cases.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -32,7 +32,7 @@ type Iface struct {
 // GetLeadingInterface ... get leading clock interface
 func (i *IFaces) GetLeadingInterface() Iface {
 	for _, iface := range *i {
-		if iface.Source == event.GNSS || iface.Source == event.PTP4l {
+		if iface.Source == event.GNSS || iface.Source == event.PTP4l || iface.Source == event.PPS {
 			return iface
 		}
 	}
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -1,0 +1,78 @@
+package config_test
+
+import (
+	"testing"
+
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/config"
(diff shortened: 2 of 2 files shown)


lack:
Closing in favor of #230

PR #229, opened 2026-06-24
Log per-port ptp4l state in UpdateState 

gtannous-spec:
## Summary
When multiple upstream ports exist (dual-TR), UpdateState() now logs per-port state and offset instead of only the aggregate, making it clear which port is stale vs active during holdover recovery.


nocturnalastro (approved):
/lgtm

PR #230, opened 2026-06-24
Fix source identification in ptpconfigs

sebsoto:
When `ts2phc.master 0` is set explicitly, the config parser would incorrectly determine that an interface had a PPS source instead of GNSS, even when a [nmea] section with `ts2phc.master 1` was present.

This commit fixes that, causing gmInterface to be set to the proper value in both the created gpsd struct, and in metrics.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interface source selection in generated timing configuration when GNSS/NMEA settings are present.
  * Ensured interfaces correctly inherit or override their source based on nearby configuration sections.
  * Added coverage for this behavior and tightened validation in existing timing-related tests.

### DIFF
--- a/pkg/daemon/testdata/profile-tgm.yaml
+++ b/pkg/daemon/testdata/profile-tgm.yaml
@@ -214,6 +214,8 @@ ptp4lOpts: -2 --summary_interval -4
 ptpSchedulingPolicy: SCHED_FIFO
 ptpSchedulingPriority: 10
 ts2phcConf: |
+  [nmea]
+  ts2phc.master 1
   [global]
   use_syslog  0
   verbose 1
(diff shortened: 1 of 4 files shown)


pkg/daemon/testdata/profile-tgm.yaml
```diff
@@ -214,22 +214,24 @@ ptp4lOpts: -2 --summary_interval -4
...
+  [nmea]
+  ts2phc.master 1
   [global]
   use_syslog  0
   verbose 1
   logging_level 7
   ts2phc.pulsewidth 100000000
   leapfile  /usr/share/zoneinfo/leap-seconds.list
   [ens7f0]
+  ts2phc.master 0
```
vitus133, line 226: looks like we don't have tis setting in the [RDS](https://github.com/openshift-kni/telco-reference/blob/9e496734099c180155b3223348da255ce0fe1a74/telco-ran/configuration/source-crs/ptp-operator/configuration/PtpConfigThreeCardGmWpc.yaml#L122). I just verified it seems to work with and without it. Worth fixing?

vitus133 (approved):
Love it!

pkg/daemon/testdata/profile-tgm.yaml
nocturnalastro: `ts2phc.master 0` is the default

PR #231, opened 2026-06-25
CNF-22693: Support dynamic namespace for OLMv1 AllNamespaces

jzding:
## Summary
Make `PtpNamespace` dynamic by reading from the `NAME_SPACE` env var, falling back to `openshift-ptp`. This is needed for OLMv1 AllNamespaces install mode where the operator may run in a different namespace.

## Changes
- `pkg/daemon/daemon.go`: `const PtpNamespace` → `var PtpNamespace` + `init()` reading `NAME_SPACE` env var

The `NAME_SPACE` env var is already passed by the ptp-operator DaemonSet template. All callers of `PtpNamespace` (leap configmap, NodePtpDevice status, PtpConfig status, pod lookup) use the variable and need no changes.

## Related PRs
- ptp-operator: [k8snetworkplumbingwg/ptp-operator#262](https://github.com/k8snetworkplumbingwg/ptp-operator/pull/262)

Fixes: [CNF-22693](https://redhat.atlassian.net/browse/CNF-22693)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * The daemon’s namespace can now be overridden at startup using an environment variable.
  * If no value is provided, the default namespace remains unchanged.

* **Bug Fixes**
  * Improved build behavior when Git metadata is not available, preventing failures in non-Git environments.

* **Chores**
  * Updated the container build process to avoid assuming local `.git` content is present.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,8 +13,8 @@ COPY deploy deploy
 COPY hack hack
 COPY Makefile Makefile
 COPY pkg pkg
-COPY .git .git
-
+ARG GIT_COMMIT=unknown
+ENV GIT_COMMIT=${GIT_COMMIT}
 RUN --mount=type=cache,target=/root/.cache/go-build make
 
 FROM quay.io/centos/centos:stream9
--- a/hack/build-image.sh
+++ b/hack/build-image.sh
@@ -6,4 +6,5 @@ IMAGE_TAG_BASE="${IMAGE_TAG_BASE:-ghcr.io/k8snetworkplumbingwg/${IMAGE_NAME}}"
 VERSION="${VERSION:-latest}"
 IMG="${IMAGE_TAG_BASE}:${VERSION}"
 
-$CONTAINER_TOOL build -t "${IMG}"  -f ./Dockerfile .
+GIT_COMMIT="${GIT_COMMIT:-$(git rev-list -1 HEAD 2>/dev/null || echo unknown)}"
+$CONTAINER_TOOL build --build-arg GIT_COMMIT="${GIT_COMMIT}" -t "${IMG}" -f ./Dockerfile .
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -3,6 +3,10 @@
 ORG_PATH="github.com/k8snetworkplumbingwg"
 REPO_PATH="${ORG_PATH}/linuxptp-daemon"
 
-GIT_COMMIT=$(git rev-list -1 HEAD)
+# GIT_COMMIT can be set via:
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -13,7 +13,7 @@ COPY deploy deploy
 COPY hack hack
 COPY Makefile Makefile
 COPY pkg pkg
-COPY .git .git
+# .git may not be present on remote builders; hack/build.sh handles this gracefully
```
nocturnalastro: This will always make it no present within the container. Is it when using a worktree that you're seeing no `.git` dir? 


Maybe something like
`COPY .gi[t] ./`
 would work as glob has a find component to its behavior

jzding: Not a worktree issue — \`ibuild\` rsyncs to a remote x86 builder with \`--exclude=.git\` (ARM Mac can't build linux/amd64 natively), so \`.git\` is never present on the remote.

I tested \`COPY .gi[t] ./\` on the remote builder (Podman 4.6.1) and it fails when \`.git\` doesn't exist:

```
STEP 2/2: COPY .gi[t] .git/
Error: copier: stat: ["/.gi[t]"]: no such file or directory
```

It works fine when \`.git\` is present, but the glob isn't treated as optional on this Podman version — it still requires at least one match.

So the fix is to remove the \`COPY .git\` line entirely and guard \`hack/build.sh\` to fall back gracefully:

```bash
if [ -z "${GIT_COMMIT:-}" ]; then
  if git rev-parse --git-dir >/dev/null 2>&1; then
    GIT_COMMIT=$(git rev-list -1 HEAD)
  else
    GIT_COMMIT="unknown"
  fi
fi
```

\`ibuild\` already injects \`GIT_COMMIT\` via \`sed\` on the remote builder before the build runs, so the version embedding still works correctly.

nocturnalastro: Whats ibuild?

jzding: sorry `ibuild` is a local wrapper for rsync I use to build binaries in remote machines. The point is `COPY .gi[t] ./` does not work with Podman 4.6.1, hence the workaround in hack/build.sh.

jzding: I simplified the code by using `ARG GIT_COMMIT` in Dockerfile. It should speed up the build as well since .git directory could be quite large and we don't really need the whole thing.

nocturnalastro: I don't think art uses this I think we're fine to change it.

nocturnalastro (approved):
/lgtm

PR #235, opened 2026-06-29
Introduce clock objects and CEPv2 IPC sending

sebsoto:
Refactors the EventHandler to manage Clock objects which contain state
decisions specific to the type of clock.

Moves state logic out of EventHandler, having it serve as a dispatcher
of events that come in, serving only relevant events to each clock.

Each clock object will calculate its new state after an event and send
any change in state as an IPC event to send to cloud event proxy v2.

Cloud event proxy 1 logic remains untouched and working. Can be removed
in a future commit.

Assisted by AI

## Summary of updates

* **New Features**
  * Added IPC caching and an IPC writer for improved, deduplicated socket message delivery.
  * Added chronyd support: parsing and event emission for chronyd-selected and no-source cases.
* **Bug Fixes**
  * Improved boundary-clock clock-class reporting and upstream dataset updates with stricter gating.
  * Per-port and per-clock handling for T-BC is now aligned with the finalized clock type.
* **Tests**
  * Expanded integration coverage for clock-class transitions, IPC behavior, and metrics (including multi-clock isolation).

### DIFF
--- a/pkg/cep/cep.go
+++ b/pkg/cep/cep.go
@@ -91,7 +91,7 @@ func buildDataValues(resourceAddr string, v ipc.Value) []event.DataValue {
 			ValueType: event.ENUMERATION,
 			Value:     val.State,
 		}
-		return []event.DataValue{dv, metricDV(resourceAddr, placeholderOffset)}
+		return []event.DataValue{dv, metricDV(resourceAddr, val.Offset)}
 	case ipc.GNSSStateValue:
 		dv := event.DataValue{
 			Resource:  resourceAddr,
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -37,6 +37,7 @@ import (
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/leap"
 
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/event"
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/ipc"
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/plugin"
 
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/logfilter"
@@ -62,6 +63,7 @@ const (
 	PTP4L_CONF_DIR                  = "/ptp4l-conf"
 	connectionRetryInterval         = 1 * time.Second
 	eventSocket                     = "/cloud-native/events.sock"
+	ipcSocket                       = "/var/run/ptp/ipc.sock"
 	ClockClassChangeIndicator       = "selected best master clock"
 	GPSDDefaultGNSSSerialPort       = "/dev/gnss0"
 	NMEASourceDisabledIndicator     = "nmea source timed out"
@@ -220,7 +222,7 @@ func (p *ProcessManager) SetTestData(name, msgTag string, ifaces config.IFaces) 
 	p.process[0].messageTag = msgTag
 	p.process[0].ifaces = ifaces
 	p.process[0].logParser = getParser(name)
-	p.process[0].handler = event.Init("test", false, eventSocket, eventChannel, closeManager, Offset, ClockState, ClockClassMetrics)
+	p.process[0].handler = event.Init("test", false, eventSocket, eventChannel, closeManager, Offset, ClockState, ClockClassMetrics, nil)
 	// Reset aliases for each test to avoid cross-case collisions.
 	alias.ClearAliases()
 	// Calculate aliases for the test interfaces to ensure proper aliasing
@@ -272,14 +274,15 @@ func (p *ProcessManager) EmitProcessStatusLogs() {
 	}
 }
 
-// EmitClockClassLogs ...
+// EmitClockClassLogs re-emits the last known clock class for each profile
+// that has an active PMC process. Called when a new consumer connects.
 func (p *ProcessManager) EmitClockClassLogs() {
 	for _, proc := range p.process {
 		if proc.name == ptp4lProcessName {
 			for _, dp := range proc.depProcess {
 				if dp.Name() == PMCProcessName {
 					pmc := dp.(*PMCProcess)
-					pmc.EmitClockClassLogs()
+					go p.ptpEventHandler.EmitClockClass(pmc.configFileName)
 				}
 			}
 		}
@@ -431,6 +434,7 @@ type Daemon struct {
 	namespace string
 	// write logs to socket, this will also send metrics to the socket
 	stdoutToSocket bool
+	enableCepV2    bool
 
 	// kubeClient allows interaction with Kubernetes, including the node we are running on.
 	kubeClient *kubernetes.Clientset
@@ -586,6 +590,7 @@ func New(
 	nodeName string,
 	namespace string,
 	stdoutToSocket bool,
+	enableCepV2 bool,
 	kubeClient *kubernetes.Clientset,
 	ptpClient *ptpclient.Clientset,
 	ptpUpdate *LinuxPTPConfUpdate,
@@ -603,12 +608,16 @@ func New(
 	InitializeOffsetMaps()
 	pluginManager, unknownPlugins := registerPlugins(plugins)
 	eventChannel := make(chan event.Event, 100)
+	cache := ipc.NewCache(100, enableCepV2)
 	pm := &ProcessManager{
 		process:         nil,
 		eventChannel:    eventChannel,
-		ptpEventHandler: event.Init(nodeName, stdoutToSocket, eventSocket, eventChannel, closeManager, Offset, ClockState, ClockClassMetrics),
+		ptpEventHandler: event.Init(nodeName, stdoutToSocket, eventSocket, eventChannel, closeManager, Offset, ClockState, ClockClassMetrics, cache),
 	}
 	tracker.processManager = pm
+	if enableCepV2 {
+		go ipc.NewWriter(ipcSocket, cache.Out(), closeManager).Run()
+	}
 
 	// Initialize fsnotify watcher for sa_file change detection
 	saFileWatch, err := fsnotify.NewWatcher()
@@ -633,6 +642,7 @@ func New(
 		nodeName:             nodeName,
 		namespace:            namespace,
 		stdoutToSocket:       stdoutToSocket,
+		enableCepV2:          enableCepV2,
 		kubeClient:           kubeClient,
 		ptpClient:            ptpClient,
 		ptpUpdate:            ptpUpdate,
@@ -784,6 +794,7 @@ func (dn *Daemon) applyNodePTPProfiles() error {
 	glog.Infof("in applyNodePTPProfiles - starting to apply %d node profiles", len(dn.ptpUpdate.NodeProfiles))
 
 	dn.stopAllProcesses()
+	dn.processManager.ptpEventHandler.RemoveAllClocks()
 	// All process should have been stopped,
 	// clear process in process manager.
 	// Assigning processManager.process to nil releases
@@ -1050,7 +1061,7 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) 
 		case TGM:
 			clockType = event.GM
 		case TBC:
-			clockType = event.BC
+			clockType = event.TBC
 			leadingNic = (*nodeProfile).PtpSettings["leadingInterface"]
 			if portsStr, ok := (*nodeProfile).PtpSettings["upstreamPort"]; ok {
 				upstreamPorts = strings.Split(portsStr, ",")
@@ -1076,7 +1087,15 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) 
 		clockType = ptp4lOutput.clock_type
 	}
 
+	clockCfgName := fmt.Sprintf("ptp4l.%d.config", runID)
+	if _, err = dn.processManager.ptpEventHandler.AddClock(clockCfgName, clockType); err != nil {
+		return fmt.Errorf("failed to register clock for profile %s: %v", *nodeProfile.Name, err)
+	}
+
 	for _, pProcess := range ptpProcesses {
+		if pProcess == chronydProcessName && !dn.enableCepV2 {
+			continue
+		}
 		controlledConfigFile := ""
 		switch pProcess {
 		case ptp4lProcessName:
@@ -1239,7 +1258,7 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) 
 		}
 
 		if pProcess == ptp4lProcessName {
-			if len(upstreamPorts) > 0 && clockType == event.BC {
+			if len(upstreamPorts) > 0 && clockType == event.TBC {
 				dprocess.tBCAttributes.trIfaceNames = upstreamPorts
 				dprocess.tBCAttributes.perPortState = make(map[string]event.PTPState, len(upstreamPorts))
 				for _, p := range upstreamPorts {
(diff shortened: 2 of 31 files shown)


pkg/daemon/daemon.go
```diff
@@ -600,12 +604,16 @@ func New(
...
 	eventChannel := make(chan event.Event, 100)
+	cache := ipc.NewCache(100, enableCepV2)
 	pm := &ProcessManager{
 		process:         nil,
 		eventChannel:    eventChannel,
-		ptpEventHandler: event.Init(nodeName, stdoutToSocket, eventSocket, eventChannel, closeManager, Offset, ClockState, ClockClassMetrics),
+		ptpEventHandler: event.Init(nodeName, stdoutToSocket, eventSocket, eventChannel, closeManager, Offset, ClockState, ClockClassMetrics, cache),
 	}
 	tracker.processManager = pm
+	if enableCepV2 {
```
nocturnalastro, line 618: Do we have a plan for how many version we will keep supporting both?

pkg/event/clock.go
```diff
@@ -0,0 +1,131 @@
...
+	emitClockClass(clockClass fbprotocol.ClockClass, cfgName string)
+	requestClockClassChange(req ClockClassRequest)
+	updateDownstreamData(bc *TBCClock, cfgName string)
+	sendIPC(msg ipc.Message)
+}
+
+// GMClock is a Grand Master clock instance.
+// It owns its sync state and emits IPC messages via clockIO.
+type GMClock struct {
+	cfgName                string
```
nocturnalastro, line 38: We could have a basic clock object which has common values which we could just embed in the clock types

pkg/event/clock.go
```diff
@@ -0,0 +1,150 @@
...
+				clockAccuracy: fbprotocol.ClockAccuracyUnknown,
+			},
+			overallSyncState:       PTP_NOTSET,
+			osClockState:           PTP_NOTSET,
+			announcedClockClass:    protocol.ClockClassUninitialized,
+			announcedClockAccuracy: fbprotocol.ClockAccuracyUnknown,
+			leadingClockData:       newLeadingClockParams(),
+		}, nil
+	case BC:
+		return &BCClock{
```
nocturnalastro, line 138: I don't see where BCClock is defined.

pkg/event/clock.go
```diff
@@ -0,0 +1,150 @@
...
+				clockClass:    protocol.ClockClassUninitialized,
+				clockAccuracy: fbprotocol.ClockAccuracyUnknown,
+			},
+			overallSyncState:       PTP_NOTSET,
+			osClockState:           PTP_NOTSET,
+			gnssState:              PTP_NOTSET,
+			announcedClockClass:    protocol.ClockClassUninitialized,
+			announcedClockAccuracy: fbprotocol.ClockAccuracyUnknown,
+		}, nil
+	case TBC:
```
nocturnalastro, line 122: Might also want T-TSC. Which is the the TR part of T-BC only.

pkg/event/event.go
```diff
@@ -651,52 +398,27 @@ func (e *EventHandler) hasMetric(name string) (*prometheus.GaugeVec, bool) {
...
+// EmitClockClass re-emits the stored clock class for any clock type.
+func (e *EventHandler) EmitClockClass(cfgName string) {
+	clk := e.GetClock(cfgName)
+	if clk == nil {
+		return
+	}
+	var clockClass fbprotocol.ClockClass
+	switch c := clk.(type) {
+	case *GMClock:
+		clockClass = c.announcedClockClass
```
nocturnalastro, line 410: Could this just be a function on the clock?

pkg/event/event.go
```diff
@@ -712,6 +434,82 @@ func (e *EventHandler) emitClockClass(clockClass fbprotocol.ClockClass, cfgName
...
 }
 
+func (e *EventHandler) sendIPC(msg ipc.Message) {
+	if e.ipcCache != nil {
+		e.ipcCache.Send(msg)
+	}
+}
+
+func (e *EventHandler) getGMSettings(cfgName string) (protocol.GrandmasterSettings, error) {
+	return pmc.GetGMSettings(cfgName)
```
nocturnalastro, line 444: What are we getting from these thin wrappers 

What about defining an interface adding it to the event handler would make it easier to mock.

pkg/event/event.go
```diff
@@ -712,6 +434,82 @@ func (e *EventHandler) emitClockClass(clockClass fbprotocol.ClockClass, cfgName
...
+func (e *EventHandler) setExternalGMPropertiesNP(cfgName string, egp protocol.ExternalGrandmasterProperties) error {
+	return pmc.SetExternalGMPropertiesNP(cfgName, egp)
+}
+
+func (e *EventHandler) getParentTimeAndCurrentDS(cfgName string) (pmc.ParentTimeCurrentDS, error) {
+	return pmc.GetParentTimeAndCurrentDS(cfgName)
+}
+
+func (e *EventHandler) getUtcOffset() int {
+	return leap.GetUtcOffset()
```
nocturnalastro, line 460: maybe leep manager should a value on event handler

pkg/event/event.go
```diff
@@ -712,6 +434,82 @@ func (e *EventHandler) emitClockClass(clockClass fbprotocol.ClockClass, cfgName
...
+
+// IPCCache returns the IPC cache, or nil if not configured.
+func (e *EventHandler) IPCCache() *ipc.Cache {
+	return e.ipcCache
+}
+
+// handleOSClockEvent fans out a PHC2SYS/CHRONYD event to all TBCClocks,
+// emits os_clock_state IPC once (system-wide), and emits sync_state IPC
+// per-profile when the overall state changes.
+func (e *EventHandler) handleOSClockEvent(event Event) {
```
nocturnalastro, line 471: We changed the logic for this fairly recently will want to update this with the E1 and E3 logic.

pkg/event/event.go
```diff
@@ -825,90 +619,93 @@ func (e *EventHandler) ProcessEvents() {
...
-		}()
-		redialClockClass = false
-	}
+		}
+	}()
 	glog.Info("starting state monitoring...")
 	for {
 		select {
 		case event := <-e.processChannel: // for non GM this thread will be in sleep forever
+			glog.V(2).Infof("ProcessEvents: received event source=%s iface=%s cfg=%s clockType=%s reset=%v",
```
nocturnalastro, line 676: I think our default V(N) is N=10 for some reason. We would still see this. Just making sure you want that.

pkg/event/event.go
```diff
@@ -936,39 +729,48 @@ func (e *EventHandler) ProcessEvents() {
...
+							if ptp, isPTP := event.Data.(*PTPData); isPTP {
+								if _, hasNMEA := ptp.Values[NMEA_STATUS]; hasNMEA {
+									ptp.Values[NMEA_STATUS] = 0
+								}
 							}
 						}
+					case *TBCClock:
+						event, clockState, dataDetails = c.addEvent(event)
+					case *BCClock:
+						dataDetails = e.addEvent(event)
```
nocturnalastro, line 763: Should we move all the values the clock object and its upto the clock to hold its values that these just becomes `c.AddEvent(event)` with no type checking? 

I understand this might be a bigger refactor job.

pkg/event/clock.go
```diff
@@ -0,0 +1,150 @@
...
+func (c *OCClock) SystemClockUpdate(_ PTPState) {}
+
+func (c *OCClock) ParentDSUpdate(_ protocol.ParentDataSet) {}
+
+// ConfigName returns the config profile name.
+func (c *OCClock) ConfigName() string { return c.cfgName }
+
+// newClock creates the appropriate Clock implementation for the given clock type.
+func newClock(cfgName string, clockType ClockType, io clockIO) (Clock, error) {
+	switch clockType {
```
nocturnalastro, line 106: We probably want a Active/Backup meta clock for HA and GM/T-BC or these might be different enough we have separate types for them.

Something which can hold other clocks. With One being active and the other/s being backups. It should proxy calls the the primary. 

This would break the type checking so we'd need add a function which just returns its clock type or have an active clock func which for most functions returns itself.

pkg/event/event.go
```diff
@@ -1369,12 +1029,18 @@ func (e *EventHandler) EmitClockSyncLogs() {
...
-		if syncState.clkLog != "" {
-			logs = append(logs, syncState.clkLog)
+	logs := make([]string, 0, len(e.clocks))
+	for _, clk := range e.clocks {
+		switch c := clk.(type) {
+		case *GMClock:
+			if c.syncState.clkLog != "" {
+				logs = append(logs, c.syncState.clkLog)
+			}
+		case *TBCClock:
```
nocturnalastro, line 1040: Can you not just do 
```go
		case *GMClock, *TBCClock:
```

pkg/event/clock.go
nocturnalastro: NVM found it. Why does it have its own file?

pkg/event/event_gm.go
```diff
@@ -0,0 +1,454 @@
...
+	fbprotocol "github.com/facebook/time/ptp/protocol"
+	"github.com/golang/glog"
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/debug"
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/ipc"
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/parser"
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/protocol"
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/utils"
+)
+
+func (c *GMClock) ParentDSUpdate(_ protocol.ParentDataSet) {}
```
nocturnalastro, line 17: Why is struct def not in here?

pkg/event/event.go
```diff
@@ -712,6 +434,82 @@ func (e *EventHandler) emitClockClass(clockClass fbprotocol.ClockClass, cfgName
 	}
 }
 
+func (e *EventHandler) sendIPC(msg ipc.Message) {
+	if e.ipcCache != nil {
+		e.ipcCache.Send(msg)
+	}
+}
+
+func (e *EventHandler) getGMSettings(cfgName string) (protocol.GrandmasterSettings, error) {
```
nocturnalastro, line 443: I think we should embbed a PMC client into the event handler which has these it would be easier to mock.

pkg/event/event_bc.go
```diff
@@ -0,0 +1,91 @@
...
+func (c *BCClock) ClockClass() fbprotocol.ClockClass {
+	return c.clockClass
+}
+
+// ConfigName returns the config profile name.
+func (c *BCClock) ConfigName() string { return c.cfgName }
+
+// BCProcessResult holds the outcome of BCClock.addEvent.
+type BCProcessResult struct {
+	State            PTPState
```
nocturnalastro, line 34: Previous state might be useful here. And a reason

nocturnalastro: Could probably make a state generic transition object. Maybe even treat is like a command object with things like updateDownstreamData, etc so each clock doesn't have to define how to do that.

pkg/event/event_tbc.go
```diff
@@ -378,47 +388,39 @@ func (e *EventHandler) announceLocalData(cfgName string) {
...
-	defer e.Unlock()
-	stateData, ok := e.clkSyncState[cfgName]
-	if !ok || stateData.state != PTP_LOCKED {
-		state := PTP_NOTSET
-		if ok {
-			state = stateData.state
-		}
-		glog.Infof("downstreamAnnounceIWF: BC state is %s (not LOCKED) %s, aborting", state, context)
+func (c *TBCClock) applyIfLockedBC(context string, fn func()) bool {
+	c.io.Lock()
```
nocturnalastro, line 403: Why is the lock on `io` and not whole clock? Isn't the at least part of the purpose for the lock that we don't want `c.syncState.state` to change during the application of the function?

pkg/event/event.go
nocturnalastro: Looks like clockIO already does this so not sure why these are here.

pkg/event/event.go
nocturnalastro: NVM clockIO does this

pkg/event/event_test_helpers_test.go
```diff
@@ -12,17 +12,28 @@ type EventHandlerForTests struct {
...
 	return &EventHandlerForTests{
-		EventHandler: newTestEventHandler(socketPath),
+		EventHandler: handler,
 	}
 }
 
-// SetClockClass populates the clock sync state for the given config
-// so that EmitClockClass has data to emit.
+// SetClockClass registers a TBC clock with the given announced clock class.
 func (t *EventHandlerForTests) SetClockClass(cfgName string, clockClass uint8) {
```
nocturnalastro, line 22: Feels like this should have TBC in the name or be a generic in some way.

pkg/ipc/cache.go
```diff
@@ -0,0 +1,117 @@
...
+	return true
+}
+
+// Snapshot returns a copy of all cached messages.
+func (c *Cache) Snapshot() []Message {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	msgs := make([]Message, 0, len(c.store))
+	for _, msg := range c.store {
+		msgs = append(msgs, msg)
```
nocturnalastro, line 74: Do we want to do a deep copy here?

nocturnalastro: Probably over complicating but perhaps this should take a function returns T and this should return []T that way it could do a deep copy or just return the encoded data out right depending on the use-case.

pkg/ipc/cache.go
```diff
@@ -0,0 +1,117 @@
...
+	defer c.mu.Unlock()
+	c.store = make(map[cacheKey]Message)
+}
+
+// Out returns the read-only outbound message channel.
+func (c *Cache) Out() <-chan Message {
+	return c.outCh
+}
+
+func valuesEqual(a, b Message) bool {
```
nocturnalastro, line 91: Maybe message should have an Equals function on it?

sebsoto:
@coderabbitai review

pkg/clockmgr/clockmgr.go
```diff
@@ -0,0 +1,377 @@
...
+	m.clocks = map[string]clock.Clock{}
+	m.clockManagementMu.Unlock()
+
+	debug.ClearState()
+}
+
+// GetClock returns the Clock registered for the given config name, or nil.
+func (m *ClockManager) GetClock(cfgName string) clock.Clock {
+	return m.clocks[cfgName]
+}
```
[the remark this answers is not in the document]
sebsoto: This isnt called from outside ProcessEvents, which does hold the lock.

pkg/cep/cep.go
nocturnalastro, line 75: Can we get rid of this now?

vitus133:
`TBC` has no mutex, but `updateDownstreamData` spawns unsynchronized goroutines (`downstreamAnnounceIWF`, `announceLocalData`) that read/write the same fields the main event loop touches, creating a data race.

`TBC` struct has no synchronization primitive:

https://github.com/k8snetworkplumbingwg/linuxptp-daemon/blob/bb8d79429834809e3c1ce804315df0e09104a7af/pkg/clock/tbc.go#L60-L74

`updateDownstreamData` fires a background goroutine per `LOCKED`/non-`LOCKED` transition that mutates `c.leadingClockData`, `c.syncState`, `c.announcedClockClass`/`c.announcedClockAccuracy` with no lock:

https://github.com/k8snetworkplumbingwg/linuxptp-daemon/blob/bb8d79429834809e3c1ce804315df0e09104a7af/pkg/clock/tbc.go#L319-L334

`ClockManager.clockManagementMu` only guards the synchronous `clk.AddEvent(ev)` call and is released immediately after, while the spawned goroutine (which performs a real PMC RPC round-trip) keeps running and can race with the next `AddEvent` call for the same clock:

https://github.com/k8snetworkplumbingwg/linuxptp-daemon/blob/bb8d79429834809e3c1ce804315df0e09104a7af/pkg/clockmgr/clockmgr.go#L215-L230

This is the same class of unsynchronized-goroutine-vs-main-loop race that reviewers previously flagged on PR #148 against the old `event_tbc.go`; it appears to have been reintroduced in this rewrite. `go test -race` is not run in CI (`hack/unit-test.sh`), so it would not currently be caught automatically, and DPLL/ts2phc events typically arrive on ~1s intervals, well within the latency of a PMC round-trip.

🤖 Generated with [opencode](https://opencode.ai)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

nocturnalastro:
> against the old `event_tbc.go`; it appears to have been reintroduced in this rewrite. `go test -race` is not run in CI (`hack/unit-test.sh`), so it would not currently be caught automatically, and DPLL/ts2phc events typically arrive on ~1s intervals, well within the latency of a PMC round-trip.

Good point we should add --race into our CI. @vitus133 I'll write a tech debt ticket for that

vitus133 (approved):

PR #239, opened 2026-07-08
CNF-25385: Add automatic interface name resolution for RHEL 10 npN suffix

jzding:
## Summary

RHEL 10 (systemd v252+) adds an `npN` suffix to multi-port NIC interface names (e.g., `ens5f0` → `ens5f0np0`). After upgrading from OCP 4.22 (RHEL 9) to 5.0 (RHEL 10), PtpConfig and HardwareConfig CRs referencing old names break because the kernel names change.

This adds an `InterfaceResolver` in `pkg/network/resolve.go` that transparently maps old names to new `npN`-suffixed names at config application time:

- Scans `/sys/class/net/` each config cycle to discover current interface names
- If a configured name doesn't exist, matches it against known interfaces using a regex pattern (`configuredName + np\d+`)
- Sorts candidates by numeric npN suffix for deterministic selection
- Resolves **PtpProfile** fields:
  - `Interface` field
  - `PtpSettings` values (`leadingInterface`, `upstreamPort`)
  - `PtpSettings` keys (`dpll.<ifname>.*`, `clockId[<ifname>]`)
  - Plugin device lists and map keys (`devices`, `pins`, `frequencies`, `phaseOffsetPins`)
- Resolves **config section headers** via `Ptp4lConf.ResolveInterfaceNames()` after parsing, reusing the existing parsed structure from `PopulatePtp4lConf`
- Resolves **HardwareConfig** fields in `UpdateHardwareConfig()` before processing:
  - `Subsystem.DPLL.NetworkInterface`
  - `Subsystem.Ethernet[].Ports[]`
  - `SourceConfig.PTPTimeReceivers[]`
- Detects key-rename collisions and skips with a warning instead of silently overwriting
- No-op when names already match (forward-compatible with RHEL 10 native configs)
- Logs all remappings at Info level; warns on refresh failures and non-string plugin device entries

**Affected NICs:** Intel E810/E825 (ice driver), WPC, GNR-D. Mellanox ConnectX already uses `npN` on RHEL 9 (unaffected).

## Test plan

- [x] 13 unit tests in `pkg/network/resolve_test.go` — name exists, npN resolution, multiple candidates, no match, empty name, caching, profile resolution (Interface, PtpSettings keys/values, plugins), HardwareConfig resolution, nil fields
- [x] 3 unit tests in `pkg/daemon/daemon_internal_test.go` — `Ptp4lConf.ResolveInterfaceNames` (resolves sections, skips known sections, no-op when names exist)
- [x] `golangci-lint run ./pkg/network/ ./pkg/daemon/ ./pkg/hardwareconfig/` — 0 issues
- [x] Integration test on OCP 5.0 node (cnfdg14) — PtpConfig with `eno12399` (no `np0` suffix) resolved to `eno12399np0` in rendered ptp4l config; ptp4l ran successfully with resolved interface
- [x] Upgrade scenario test on OCP 4.22 node (cnfdg32) — PtpConfig with `ens6f0` (old-style name) resolved to `ens6f0np0` (actual system interface); ptp4l bound to correct interface. Existing T-GM config with `ens7f0`/`ens7f1` passed through unchanged (resolver no-op). Original T-GM sync restored after cleanup.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added automatic network interface name resolution for PTP profiles, hardware configurations, and generated PTP configuration.
  * Supports remapping interface references across clock-chain settings and plugin configurations.
  * Selects matching system interfaces deterministically and refreshes mappings when configurations are applied.

* **Bug Fixes**
  * Prevents configuration failures when configured interface names differ from the system’s actual names.

* **Tests**
  * Added coverage for interface mapping across profiles, hardware configurations, plugins, and PTP configuration sections.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -465,6 +465,8 @@ type Daemon struct {
 
 	delayedPhc2sys   atomic.Bool
 	delayedPhc2sysMu sync.Mutex // protects skipInitialStartup on phc2sys processes
+
+	interfaceResolver *ptpnetwork.InterfaceResolver
 }
 
 type initialStateSyncer interface{ SyncInitialState() }
@@ -626,24 +628,25 @@ func New(
 	}
 
 	dn := &Daemon{
-		nodeName:              nodeName,
-		namespace:             namespace,
-		stdoutToSocket:        stdoutToSocket,
-		kubeClient:            kubeClient,
-		ptpClient:             ptpClient,
-		ptpUpdate:             ptpUpdate,
-		pluginManager:         pluginManager,
-		unknownPlugins:        unknownPlugins,
-		hwconfigs:             hwconfigs,
-		hardwareConfigManager: hardwareconfig.NewHardwareConfigManager(kubeClient, namespace),
-		refreshNodePtpDevice:  refreshNodePtpDevice,
-		pmcPollInterval:       pmcPollInterval,
-		processManager:        pm,
-		readyTracker:          tracker,
-		stopCh:                stopCh,
-		saFileWatcher:         saFileWatch,
-		liveGate:              &liveGate{},
+		nodeName:             nodeName,
+		namespace:            namespace,
+		stdoutToSocket:       stdoutToSocket,
+		kubeClient:           kubeClient,
+		ptpClient:            ptpClient,
+		ptpUpdate:            ptpUpdate,
+		pluginManager:        pluginManager,
+		unknownPlugins:       unknownPlugins,
+		hwconfigs:            hwconfigs,
+		interfaceResolver:    ptpnetwork.NewInterfaceResolver(),
+		refreshNodePtpDevice: refreshNodePtpDevice,
+		pmcPollInterval:      pmcPollInterval,
+		processManager:       pm,
+		readyTracker:         tracker,
+		stopCh:               stopCh,
+		saFileWatcher:        saFileWatch,
+		liveGate:             &liveGate{},
 	}
+	dn.hardwareConfigManager = hardwareconfig.NewHardwareConfigManager(kubeClient, namespace, dn.interfaceResolver)
 	pm.daemon = dn
 	return dn
 }
@@ -833,6 +836,11 @@ func (dn *Daemon) applyNodePTPProfiles() error {
 		dn.hardwareConfigManager.SetPtpConfig(ptpConfig)
 		glog.Infof("Updated PtpConfig in hardware config manager with %d profiles (after sorting and reconciliation)", len(dn.ptpUpdate.NodeProfiles))
 	}
+	// Refresh interface resolver so RHEL 10 npN name mapping uses current system state
+	if err := dn.interfaceResolver.Refresh(); err != nil {
+		glog.Warningf("Failed to refresh interface resolver, name resolution may use stale data: %v", err)
+	}
+
 	// TODO: resolve clock IDs, clockType, leadingInterface and upstreamPort from hardware config
 	// (needed to keep code compatibility elsewhere and allow it to work both with hardware config and plugins)
 	for _, profile := range dn.ptpUpdate.NodeProfiles {
@@ -859,6 +867,8 @@ func (dn *Daemon) applyNodePTPProfiles() error {
 		if controlledID, ok := relations[*profile.Name]; ok {
 			profile.PtpSettings["controlledId"] = strconv.Itoa(controlledID)
 		}
+
+		dn.interfaceResolver.ResolveProfileInterfaces(&profile)
 
 		glog.Infof("Calling applyNodePtpProfile for profile %s with runID %d", *profile.Name, runID)
 		err := dn.applyNodePtpProfile(runID, &profile)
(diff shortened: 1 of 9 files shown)


pkg/network/resolve.go
```diff
@@ -0,0 +1,329 @@
...
+	resolveConfSections(r, profile.Synce4lConf, profileName, "synce4lConf")
+
+	if profile.PtpSettings != nil {
+		resolvePtpSettings(r, profile.PtpSettings, profileName)
+	}
+
+	resolvePluginDevices(r, profile, profileName)
+}
+
+func resolveConfSections(r *InterfaceResolver, conf *string, profileName, confField string) {
```
sebsoto: This is implementing parsing of the ptp conf, which is already defined within https://github.com/k8snetworkplumbingwg/linuxptp-daemon/blob/b51bb91d1dcf49c06e4fc8574ede56d13960b502/pkg/daemon/config.go#L238

This makes things fragile, that if the configuration changes over time, we need to change both places. It also means the effort is being made to parse it multiple times. Wouldn't it make more sense to do the appropriate replacing in applyNodePtpProfile, after the config has been fully parsed?

sebsoto (dismissed):

pkg/network/resolve.go
jzding: Good point. Added Ptp4lConf.ResolveInterfaceNames(resolver) in config.go that operates on the already-parsed sections after PopulatePtp4lConf, called in applyNodePtpProfile before RenderPtp4lConf. No duplicate parsing.

pkg/network/resolve.go
```diff
@@ -0,0 +1,295 @@
...
+		return -1
+	}
+	n := 0
+	for _, c := range iface[idx+2:] {
+		if c < '0' || c > '9' {
+			return -1
+		}
+		n = n*10 + int(c-'0')
+	}
+	return n
```
nocturnalastro, line 65: Actually why not just do `strconv.Atoi(iface[idx+2:])` and `return -1` on an error? Which is what this cirremt loop does effectively I think?

This isn't going to be called that often so I think it's ok to give up the zero-copy nature of the loop for readability.

pkg/network/resolve.go
```diff
@@ -0,0 +1,295 @@
...
+			continue
+		}
+
+		changed := false
+
+		// Resolve "devices" array
+		if devices, ok := rawOpts["devices"].([]any); ok {
+			for i, d := range devices {
+				name, isStr := d.(string)
+				if isStr {
```
nocturnalastro: And if not? we should raise a warning I think.

nocturnalastro (commented):
Do we also need hardware config support for this?

pkg/network/resolve.go
jzding: done

jzding:
> Do we also need hardware config support for this?

Added

pkg/daemon/daemon.go
```diff
@@ -627,24 +628,25 @@ func New(
...
+	}
+	dn.hardwareConfigManager = hardwareconfig.NewHardwareConfigManager(kubeClient, namespace, dn.interfaceResolver)
+	dn.refreshNodePtpDevice = refreshNodePtpDevice
+	dn.pmcPollInterval = pmcPollInterval
+	dn.processManager = pm
+	dn.readyTracker = tracker
+	dn.stopCh = stopCh
+	dn.saFileWatcher = saFileWatch
+	dn.liveGate = &liveGate{}
 	pm.daemon = dn
```
nocturnalastro: It would be neater to just create the resolver ahead of time then add it in
```suggestion
		interfaceResolver: resolver
		hardwareConfigManager: hardwareconfig.NewHardwareConfigManager(kubeClient, namespace, resolver)
	   refreshNodePtpDevice: refreshNodePtpDevice
	   pmcPollInterval: pmcPollInterval
	   processManager: pm
	   readyTracker: tracker
	   stopCh: stopCh
	   saFileWatcher: saFileWatch
	   liveGate: &liveGate{}
}
pm.daemon = dn
```

nocturnalastro: Or just pull the one value out of the constructor
```suggestion
		interfaceResolver: ptpnetwork.NewInterfaceResolver(),
	    refreshNodePtpDevice: refreshNodePtpDevice
	    pmcPollInterval: pmcPollInterval
	    processManager: pm
	    readyTracker: tracker
	    stopCh: stopCh
	    saFileWatcher: saFileWatch
	    liveGate: &liveGate{}		
	}
	dn.hardwareConfigManager = hardwareconfig.NewHardwareConfigManager(kubeClient, namespace, dn.interfaceResolver)
pm.daemon = dn
```

jzding: > Or just pull the one value out of the constructor

Done — moved everything back into the struct literal, only `hardwareConfigManager` is assigned separately since it references `dn.interfaceResolver`.

nocturnalastro (approved):
/lgtm

PR #241, opened 2026-07-09
Refactor T-BC port-state detection to use structured parser events instead of raw string matching

gtannous-spec:
Addressing tech-debt : [CNF-25298](https://redhat.atlassian.net/browse/CNF-25298)

## Summary 
Replaces ad hoc strings.Contains matching on raw ptp4l log lines with the shared, structured parser.PTPEvent (adding a new PreviousRole field) across the legacy T-BC path, the metrics fallback extractor, and the hardwareconfig state detector. Fixes a confirmed regression where the legacy path could spuriously re-enter HOLDOVER on benign, non-transition ptp4l log lines, and adds comprehensive test coverage for port role/state transitions to guard against future regressions.

## Summary by CodeRabbit

* **New Features**
  * Added a new Prometheus `ServoState` gauge (`ptp/servo_state`) and propagated linuxptp/phc2sys/ts2phc `s0–s3` parsing end-to-end.
  * Enhanced PTP event parsing to include interface name and previous role.

* **Bug Fixes**
  * Reduced false “lost sync”/holdover triggers by switching to structured ptp4l event parsing and ignoring malformed/unrelated lines.
  * Improved precision for legacy and hardware-config state-change detection.
  * Ensured servo state resets appropriately when the port transitions to faulty.

* **Tests**
  * Expanded regression and behavior-pinning tests covering holdover precision, previous-role parsing, servo-state updates, and faulty reset behavior.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1732,18 +1732,9 @@ func (p *ptpProcess) processTBCTransitionHardwareConfig(output string) {
 	})
 }
 
-// tbcPTP4LExtractor is a shared ptp4l log-line extractor used by the legacy T-BC
-// path to turn raw ptp4l output into a structured parser.PTPEvent, instead of
-// matching substrings directly against the output. Reused across calls since it
-// holds only compiled regexes and is safe for concurrent read-only use.
-var tbcPTP4LExtractor = parser.NewPTP4LExtractor()
-
-// processTBCTransitionLegacy is the original implementation as ultimate fallback.
-// It derives port role transitions from the structured parser.PTPEvent produced
-// by the shared ptp4l parser (the same one used elsewhere in the log-processing
-// pipeline), rather than performing its own ad hoc string matching on the raw
-// ptp4l output.
+// processTBCTransitionLegacy is the ultimate T-BC fallback: it derives port role transitions from the structured parser.PTPEvent instead of matching substrings on raw ptp4l output.
 func (p *ptpProcess) processTBCTransitionLegacy(output string, pm *plugin.PluginManager) {
+	tbcPTP4LExtractor := parser.NewPTP4LExtractor()
 	_, ptpEvent, err := tbcPTP4LExtractor.Extract(output)
 	if err != nil {
 		glog.Warningf("T-BC legacy: failed to parse ptp4l port event %q: %v", output, err)
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1282,7 +1282,18 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) 
 				if !clockTypeFound {
 					pmcClockType = string(clockType)
 				}
-				pmcProcess := NewPMCProcess(runID, dn.processManager.ptpEventHandler, pmcClockType)
+				var portStateOpts *PMCPortStateOpts
+				if pmcClockType == TBC && len(ifaces) > 0 {
+					portIfaceMap := make(map[int]string, len(ifaces))
+					for idx, ifc := range ifaces {
+						portIfaceMap[idx+1] = ifc.Name
+					}
+					portStateOpts = &PMCPortStateOpts{
+						PortIfaceMap: portIfaceMap,
+						OnChange:     dprocess.pmcPortStateChanged,
+					}
+				}
+				pmcProcess := NewPMCProcess(runID, dn.processManager.ptpEventHandler, pmcClockType, portStateOpts)
 				pmcProcess.CmdInit()
 				// TODO addScheduling
 				dprocess.depProcess = append(dprocess.depProcess, pmcProcess)
@@ -1763,6 +1774,16 @@ func (p *ptpProcess) processTBCTransitionLegacy(output string, pm *plugin.Plugin
 		pm.AfterRunPTPCommand(&p.nodeProfile, "tbc-ho-exit")
 		p.sendPtp4lEvent()
 	})
+}
+
+// pmcPortStateChanged is the callback invoked by PMCProcess.handlePortDS when
+// a PORT_DATA_SET notification indicates a port role change. In Phase 1 this
+// logs the derived condition for comparison with the log-scraping path;
+// the actual T-BC state machine is still driven by log scraping.
+func (p *ptpProcess) pmcPortStateChanged(iface string, role, previousRole parserconstants.PTPPortRole) {
+	_, conditionType := hardwareconfig.DetectStateChangeFromRole(iface, role, previousRole)
+	glog.Infof("PMC port-state: iface=%s role=%v previousRole=%v condition=%q (config=%s)",
+		iface, role, previousRole, conditionType, p.configName)
 }
 
 // cmdRun runs given ptpProcess and restarts on errors
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -25,6 +25,7 @@ import (
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/hardwareconfig"
 	ptpnetwork "github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/network"
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/parser"
+	parserconstants "github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/parser/constants"
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/synce"
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/ublox"
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/utils"
@@ -1731,30 +1732,33 @@ func (p *ptpProcess) processTBCTransitionHardwareConfig(output string) {
 	})
 }
 
-// processTBCTransitionLegacy is the original implementation as ultimate fallback
+// tbcPTP4LExtractor is a shared ptp4l log-line extractor used by the legacy T-BC
+// path to turn raw ptp4l output into a structured parser.PTPEvent, instead of
+// matching substrings directly against the output. Reused across calls since it
+// holds only compiled regexes and is safe for concurrent read-only use.
+var tbcPTP4LExtractor = parser.NewPTP4LExtractor()
+
+// processTBCTransitionLegacy is the original implementation as ultimate fallback.
+// It derives port role transitions from the structured parser.PTPEvent produced
+// by the shared ptp4l parser (the same one used elsewhere in the log-processing
+// pipeline), rather than performing its own ad hoc string matching on the raw
+// ptp4l output.
 func (p *ptpProcess) processTBCTransitionLegacy(output string, pm *plugin.PluginManager) {
-	portMatched := false
-	for _, iface := range p.tBCAttributes.trIfaceNames {
-		if strings.Contains(output, iface) {
-			portMatched = true
-			break
-		}
+	_, ptpEvent, err := tbcPTP4LExtractor.Extract(output)
+	if err != nil {
+		glog.Warningf("T-BC legacy: failed to parse ptp4l port event %q: %v", output, err)
 	}
-	if portMatched {
-		if strings.Contains(output, "to SLAVE on MASTER_CLOCK_SELECTED") {
-			portName := parser.ExtractPortName(output)
-			if portName != "" {
-				if len(p.tBCAttributes.trIfaceNames) > 1 && !slices.Contains(p.tBCAttributes.trIfaceNames, portName) {
-					glog.Warningf("Ignoring non-TR port in legacy MASTER_CLOCK_SELECTED event: %s", portName)
-				} else {
-					p.tBCAttributes.activePort = portName
-					glog.Infof("T-BC active upstream port: %s", portName)
-				}
-			}
+
+	if ptpEvent != nil && ptpEvent.Iface != "" && slices.Contains(p.tBCAttributes.trIfaceNames, ptpEvent.Iface) {
+		switch {
+		case ptpEvent.Role == parserconstants.PortRoleSlave:
+			// Port became (or remains) the active upstream time-receiver port.
+			p.tBCAttributes.activePort = ptpEvent.Iface
+			glog.Infof("T-BC active upstream port: %s", ptpEvent.Iface)
 			p.tBCAttributes.lastReportedState = event.PTP_LOCKED
 			p.tBCAttributes.offsetFilter = utils.NewWindow(offsetFilterSize)
-		} else if strings.Contains(output, "to MASTER on ANNOUNCE_RECEIPT_TIMEOUT_EXPIRES") ||
-			strings.Contains(output, "SLAVE to") {
+		case ptpEvent.PreviousRole == parserconstants.PortRoleSlave:
+			// Port is no longer a slave indicating it lost sync
 			pm.AfterRunPTPCommand(&p.nodeProfile, "tbc-ho-entry")
 			p.tBCAttributes.lastReportedState = event.PTP_FREERUN
 			glog.Info("T-BC MOVE TO HOLDOVER")
--- a/pkg/daemon/metrics.go
+++ b/pkg/daemon/metrics.go
@@ -138,6 +138,18 @@ var (
 			Help:      "0 = FREERUN, 1 = LOCKED, 2 = HOLDOVER",
 		}, []string{"process", "node", "iface"})
 
+	// ServoState metrics to show the raw linuxptp servo state (s0-s3) as
+	// printed in ptp4l/phc2sys/ts2phc offset log lines (e.g. "master offset
+	// -1 s2 freq +1 path delay 18481"), independent of the collapsed
+	// ClockState metric above.
+	ServoState = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: PTPNamespace,
+			Subsystem: PTPSubsystem,
+			Name:      "servo_state",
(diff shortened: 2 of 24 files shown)


pkg/parser/model.go
```diff
@@ -10,11 +10,17 @@ type StatusMetric struct {
...
-	Raw        string                `json:"raw"`        // original line
+	PortID int                   `json:"portid"`
+	Iface  string                `json:"iface"`
+	Role   constants.PTPPortRole `json:"role"` // e.g. SLAVE, MASTER, FAULTY
+	// PreviousRole is the role the port held immediately before this
+	// transition (e.g. "SLAVE to MASTER ..." -> PortRoleSlave), parsed
+	// directly from the log line's own "<FROM> to <TO>" text. It is
+	// constants.PortRoleUnknown when the line has no explicit "X to Y"
+	// transition (e.g. FAULT_DETECTED) or the FROM state isn't recognized.
+	PreviousRole constants.PTPPortRole `json:"previousrole"`
```
sebsoto, line 21: Not sure this makes sense to me. Keeping track of the past state of the system really shouldn't be done within an event.  This is duplicating data from the previous event. Rather the consumer, `ptpProcess` should track this data for each port if it is necessary to resolve the overall system state.

sebsoto (changes requested):

pkg/parser/ptp4l_parser_test.go
```diff
@@ -99,208 +99,287 @@ func TestPTP4LEventParser(t *testing.T) {
...
 			expectedEvent: &parser.PTPEvent{
-				PortID:     0,
-				Role:       constants.PortRoleUnknown,
-				ClockState: constants.ClockStateFreeRun,
-				Raw:        "ptp4l[4268779.809]: [ptp4l.0.config] port 1: INVALID_STATE",
+				PortID:       0,
+				Role:         constants.PortRoleUnknown,
+				PreviousRole: constants.PortRoleUnknown,
+				ClockState:   constants.ClockStateFreeRun,
+				Raw:          "ptp4l[4268779.809]: [ptp4l.0.config] port 1: INVALID_STATE",
```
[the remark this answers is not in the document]
gtannous-spec: Fixed.

pkg/daemon/metrics.go
```diff
@@ -654,53 +655,26 @@ func deleteProcessStatusMetrics(config, process string) {
...
-	portId, e = strconv.Atoi(portIndex)
-	if e != nil {
-		glog.Errorf("error parsing port id %s", e)
-		portId = 0
-		return
+// extractPTP4lEventState parses a ptp4l port state change log line and returns
+// the port ID and role, based on the structured parser.PTPEvent produced by the
+// shared ptp4l parser (the same one used elsewhere in the log-processing
+// pipeline), rather than performing ad hoc string matching on the raw output.
+func extractPTP4lEventState(output string) (portId int, role ptpPortRole) {
```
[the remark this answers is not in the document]
gtannous-spec: Fixed.

pkg/parser/model.go
gtannous-spec: this field exists because Role alone only tells you the destination state of a ptp4l transition. it can't tell the consumer if this port has just leave SLAVE which is exactly what the T-BC state machine needs as an answer to decide whether to declare HOLDOVER.. 

PreviousRole is the FROM side of that same "<FROM> to <TO>" text, parsed once, directly from the log line itself, with no need for the caller to remember anything across calls.

it mostly matters for hardwareconfig paths especially for `DetectStateChange` which is a stateless function, it only sees at a time with no memory of what port has been doing before..

if that name is misleading, we could rename it to something like FromRole ...

pkg/daemon/daemon.go
```diff
@@ -1701,30 +1702,40 @@ func (p *ptpProcess) processTBCTransitionHardwareConfig(output string) {
 	})
 }
 
-// processTBCTransitionLegacy is the original implementation as ultimate fallback
+// tbcPTP4LExtractor is a shared ptp4l log-line extractor used by the legacy T-BC
+// path to turn raw ptp4l output into a structured parser.PTPEvent, instead of
+// matching substrings directly against the output. Reused across calls since it
+// holds only compiled regexes and is safe for concurrent read-only use.
+var tbcPTP4LExtractor = parser.NewPTP4LExtractor()
```
sebsoto: Does this need to be a global? Looking at the function it doesn't seem there needs to be state held across calls. It also doesn't seem expensive to initialize.

pkg/parser/model.go
sebsoto: Nope that makes sense. My fault for reading through this too fast :)

pkg/daemon/metrics.go
```diff
@@ -654,53 +655,26 @@ func deleteProcessStatusMetrics(config, process string) {
...
-		glog.Errorf("failed to parse output %s: unexpected number of fields", output)
-		return
-	}
-
-	portIndex := fields[1]
-	role = UNKNOWN
+// ptp4lEventStateExtractor is the shared ptp4l parser used to derive port role
+// information from raw ptp4l output. Reused across calls since it only holds
+// compiled regexes and is safe for concurrent read-only use.
+var ptp4lEventStateExtractor = parser.NewPTP4LExtractor()
```
sebsoto: same as above

pkg/daemon/daemon.go
```diff
@@ -1701,30 +1702,40 @@ func (p *ptpProcess) processTBCTransitionHardwareConfig(output string) {
...
-			strings.Contains(output, "SLAVE to") {
+		case ptpEvent.PreviousRole == parserconstants.PortRoleSlave:
+			// This port's own log line states it was SLAVE immediately
+			// before this transition and is no longer, i.e. it just lost
+			// sync (regardless of what it became). Deliberately NOT keyed
+			// off ptpEvent.Role or ptpEvent.Iface == activePort alone: Role
+			// is populated for every parseable per-port ptp4l line (even
+			// PortRoleUnknown ones like "delay timeout"), and activePort
+			// equality is true for any line naming that port -- either
+			// would fire on unrelated, non-transition log lines.
```
sebsoto: nit: this is a big comment, couldn't it just be something like:
`// Port is no longer a slave indicating it lost sync`

I'm not sure the extra information is helpful for understanding the branching. Its good information, but I think moving it to the commit message might be more appropriate.

sebsoto (changes requested):
Thanks, changes make sense to me!

pkg/daemon/daemon.go
gtannous-spec: Thanks for the note ! you're absolutely right, we can initialize it locally inside each function where it's called.

pkg/daemon/daemon.go
gtannous-spec: Fixed.

pkg/daemon/metrics.go
gtannous-spec: @sebsoto PTAL at the new changes

vitus133:
Can we also gave ptp4l servo state extracted from logs and exposed to users?
For example
```
ptp4l[2330.162]: [ptp4l.1.config:6] master offset         -1 s2 freq      +1 path delay     18481
```
This `s2` could be s0, s1, s2, s3. Let's expose it as servo state

gtannous-spec:
> Can we also gave ptp4l servo state extracted from logs and exposed to users?

Ok will do that in a different commit 👍  @vitus133 

Edit: Done ✅

pkg/daemon/metrics.go
```diff
@@ -654,53 +655,19 @@ func deleteProcessStatusMetrics(config, process string) {
...
-	var e error
-	portId, e = strconv.Atoi(portIndex)
-	if e != nil {
-		glog.Errorf("error parsing port id %s", e)
-		portId = 0
-		return
+// extractPTP4lEventState parses a ptp4l port state change log line into a port ID and role using the structured parser.PTPEvent instead of ad hoc string matching on the raw output.
+func extractPTP4lEventState(output string) (portID int, role ptpPortRole) {
+	ptp4lEventStateExtractor := parser.NewPTP4LExtractor()
+	_, ptpEvent, err := ptp4lEventStateExtractor.Extract(output)
```
nocturnalastro, line 715: Do we have too keep extracting? Would it possible to pass the ptpEvent around?

vitus133: Ideally, why don't we stop ptp4l log scraping and start relying exclusively on pmc event monitoring?

nocturnalastro: It would be fairly easy to add that to the current PMC implementation :)

gtannous-spec: do you mean we get the Port state based on the PMC events?

gtannous-spec: In fact, PMC and ptp4l's stdout logs aren't redundant, they expose genuinely different information.
PMC cannot tell the servo state, so "stop ptp4l log scraping, rely exclusively on PMC" would mean losing servo-state visibility entirely.. 
WDYT ? @vitus133 @nocturnalastro

nocturnalastro: Port log lines don't have servo state.

nocturnalastro: I think Vitaly is just suggesting for the port logs, not for the metric logs.
At least as I understood it.

gtannous-spec: Are we talking only about PORT STATES PARSING? or do we refer also to Vitaly's comment [HERE](https://github.com/k8snetworkplumbingwg/linuxptp-daemon/pull/241#issuecomment-4957657620)

gtannous-spec: Anyways, should we skip the PMC CLI text parsing entirely by talking directly to ptp4l's management UDS using a PTP management client ?
instead of spawning the PMC CLI and regex-matching its text output.. 
@vitus133 @nocturnalastro

sebsoto (dismissed):
Thanks, my initial comments were addressed.
Will leave to others to merge, once their comments are addressed.

pkg/daemon/metrics.go
```diff
@@ -138,6 +139,18 @@ var (
 			Help:      "0 = FREERUN, 1 = LOCKED, 2 = HOLDOVER",
 		}, []string{"process", "node", "iface"})
 
+	// ServoState metrics to show the raw linuxptp servo state (s0-s3) as
+	// printed in ptp4l/phc2sys/ts2phc offset log lines (e.g. "master offset
+	// -1 s2 freq +1 path delay 18481"), independent of the collapsed
+	// ClockState metric above.
+	ServoState = prometheus.NewGaugeVec(
```
sebsoto, line 145: If my understanding is correct, this is adding to the API and needs to be documented for customers https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/advanced_networking/using-ptp-hardware#nw-ptp-operator-metrics-reference-v2_ptp-consumer

gtannous-spec: This is part of the changes Vitaly requested ! thought about doing them on a different PR, but it looks like there are multiple api-level changes

pkg/daemon/metrics.go
nocturnalastro: This thread is attached to a line of code which is just dealing with the event (the second returned values). This is just port states.

Metrics which are the first value returned, deal with the offsets, servo states etc...

Talking directly to the UDS would be an improvement to reduce string parsing. Is there a library in go you think we should use? Or are you suggesting we write our own? It may well be more effort than the string parsing. :)

pkg/daemon/metrics.go
nocturnalastro: @vitus133 isn't our locked state metric not derived from the servo state? Whats the added benefit of the raw value?

pkg/daemon/metrics.go
gtannous-spec: If that is gonna complicate things up, then let's just stick to parser..
in fact The pkg/pmc/client.go file already exists as a start in that direction but it is requires a separate larger refactor...

nocturnalastro: It would be easy to extend the string parsing to handle the port state events in the our pmc process :)

gtannous-spec: Yep ! will push a new commit for it

gtannous-spec: Hold on with the review gotta test it more

pkg/protocol/type.go
```diff
@@ -520,6 +543,140 @@ func (tp *TimePropertiesDS) String() string {
...
+		p.AnnounceReceiptTimeout == other.AnnounceReceiptTimeout &&
+		p.LogSyncInterval == other.LogSyncInterval &&
+		p.DelayMechanism == other.DelayMechanism &&
+		p.LogMinPdelayReqInterval == other.LogMinPdelayReqInterval &&
+		p.VersionNumber == other.VersionNumber
+}
+
+// PortNumber extracts the port number from the trailing "-N" of PortIdentity
+// (e.g. "b4e9b8.fffe.5ec71a-3" -> 3).
+func (p *PortDataSet) PortNumber() (int, error) {
```
nocturnalastro: We could extract this while parsing the port identity.

pkg/protocol/type.go
```diff
@@ -18,6 +18,10 @@ const (
 	ClockClassOutOfSpec     protocol.ClockClass = 140
 )
 
+// signedIntRegEx matches an optionally negative base-10 integer, shared by the
+// several DataSet ValueRegEx maps whose fields may report negative values.
+const signedIntRegEx = `-?\d+`
```
nocturnalastro: I don't think this is so complex it needs to be named.

pkg/parser/port_state.go
```diff
@@ -0,0 +1,29 @@
...
+// PortStateToRole maps a linuxptp ps_str port-state name (as carried in a
+// PORT_DATA_SET management message's portState field) to the internal
+// PTPPortRole and ClockState enums.
+//
+// The mapping mirrors determineRole() in ptp4l_parser.go but operates on a
+// single-state string ("SLAVE") rather than a log-line transition phrase
+// ("UNCALIBRATED to SLAVE on ..."). Because PORT_DATA_SET only reports the
+// current state (not the previous one), the caller is responsible for
+// maintaining previous-role context externally.
+func PortStateToRole(state string) (constants.PTPPortRole, constants.ClockState) {
```
nocturnalastro: clock state isn't used in the actual code just the test and it mostly returns freerun this miss-leading at best. Remove the second return value.

pkg/protocol/type.go
gtannous-spec: yeah surely.. but this will cause into linter errors :)

pkg/protocol/type.go
gtannous-spec: fixed.

pkg/parser/port_state.go
gtannous-spec: fixed.

pkg/daemon/pmc.go
```diff
@@ -23,9 +25,20 @@ const (
...
 
+// PortStateChangeCallback is called by PMCProcess when a port-state
+// notification is received and the role has actually changed.
+type PortStateChangeCallback func(iface string, role, previousRole constants.PTPPortRole)
+
+// PMCPortStateOpts groups the optional port-state monitoring parameters
+// for NewPMCProcess. When nil / zero-value, port-state monitoring is disabled.
+type PMCPortStateOpts struct {
+	PortIfaceMap map[int]string
+	OnChange     PortStateChangeCallback
```
nocturnalastro, line 36: Why does this need to be an injected function?

pkg/daemon/pmc.go
```diff
@@ -23,9 +25,20 @@ const (
...
 )
 
+// PortStateChangeCallback is called by PMCProcess when a port-state
+// notification is received and the role has actually changed.
+type PortStateChangeCallback func(iface string, role, previousRole constants.PTPPortRole)
+
+// PMCPortStateOpts groups the optional port-state monitoring parameters
+// for NewPMCProcess. When nil / zero-value, port-state monitoring is disabled.
+type PMCPortStateOpts struct {
+	PortIfaceMap map[int]string
```
nocturnalastro, line 35: Why do we need the interfaces?

pkg/daemon/pmc.go
```diff
@@ -47,12 +68,18 @@ type PMCProcess struct {
...
 	parentDS          *protocol.ParentDataSet
 	parentDSCh        chan protocol.ParentDataSet
+	portDSCh          chan protocol.PortDataSet
 	exitCh            chan struct{}
 	clockType         string
 	c                 net.Conn // guarded by lock
 	messageTag        string
 	eventHandler      *event.EventHandler
 
+	// Port-state monitoring fields (non-nil only when monitorPortState is true).
```
nocturnalastro, line 78: Why wouldn't we just always have this on?

pkg/daemon/pmc.go
gtannous-spec: the callback is because PMCProcess lives in pkg/daemon but has no direct reference to the ptpProcess that owns it.

pkg/daemon/pmc.go
gtannous-spec: to map ptp4l port numbers to linux iface names.. 
It's needed because PORT_DATA_SET notifications from ptp4l identify ports by their portIdentity

pkg/daemon/pmc.go
gtannous-spec: cuz it only makes sense for T-BC?

gtannous-spec: The reason it is conditional because:
The callback needs a ptpProcess to call back into. The OnChange closure captures dprocess.pmcPortStateChanged, which accesses **T-BC-specific fields,** for non TBC processes those fields aren't populated, so calling the callback would be meaningless.

nocturnalastro (approved):
/lgtm

PR #242, opened 2026-07-09
Fix AnnounceClockClass passing ClockClass as accuracy argument

sebsoto:
handleParentDS was passing GrandmasterClockClass for both the clockClass and clockAccuracy parameters causing an accuracy value equal to the clock class, instead of the actual value

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected clock announcement data so clock accuracy is now reported from the proper source, improving the accuracy of clock-class updates sent to the system.


nocturnalastro (approved):
/lgtm

PR #243, opened 2026-07-12
OCPBUGS-98473: Dell XR8720t A0 PTP delays are not calibrated

vitus133:
Fix time error by calibrating phase delays on P-Neo
Reduce log noise
Fix DPLL decoding error

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved DPLL status handling for signed fractional frequency offsets, including large values and multiple kernel encoding formats.
  - Updated timing compensation values for supported Dell XR8720t hardware to improve synchronization accuracy.
  - Reduced unnecessary status-table updates when DPLL lock status has not changed.

- **Diagnostics**
  - Moved detailed PTP, DPLL, and boundary-clock processing messages to verbose logging, reducing routine log noise while preserving troubleshooting details.

### DIFF
--- a/pkg/daemon/ptpdev.go
+++ b/pkg/daemon/ptpdev.go
@@ -107,7 +107,7 @@ func GetDevStatusUpdate(nodePTPDev *ptpv1.NodePtpDevice) (*ptpv1.NodePtpDevice, 
 		if dev.Name == nicDefaultPort[nicBase] {
 			ptpnetwork.LogStructuredHardwareInfo(dev.Name, dev.HardwareInfo)
 		} else {
-			glog.Infof("PTP Device: %s (PCI: %s, same NIC as %s)", dev.Name, dev.HardwareInfo.PCIAddress, nicDefaultPort[nicBase])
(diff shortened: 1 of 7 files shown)


pkg/daemon/ptpdev.go
```diff
@@ -107,7 +107,7 @@ func GetDevStatusUpdate(nodePTPDev *ptpv1.NodePtpDevice) (*ptpv1.NodePtpDevice,
 		if dev.Name == nicDefaultPort[nicBase] {
 			ptpnetwork.LogStructuredHardwareInfo(dev.Name, dev.HardwareInfo)
 		} else {
-			glog.Infof("PTP Device: %s (PCI: %s, same NIC as %s)", dev.Name, dev.HardwareInfo.PCIAddress, nicDefaultPort[nicBase])
+			glog.V(14).Infof("PTP Device: %s (PCI: %s, same NIC as %s)", dev.Name, dev.HardwareInfo.PCIAddress, nicDefaultPort[nicBase])
```
jzding, line 110: nit: use a constant instead of 14?

vitus133: I wonder - should we handle logging in our own logging customization package? Because otherwise we'll need to declare these constants in every package.
 I think it's more for a tech debt

lack (approved):

PR #244, opened 2026-07-13
OCPBUGS-98868: Fix stale clockClass 6 announcement to downstream ptp4l during T-BC holdover

gtannous-spec:
## Summary
The gmIdentity regex only matched decimal digits, but real PTP clock identities are hex - so every EXTERNAL_GRANDMASTER_PROPERTIES_NP PMC call silently timed out after 2s, every time. That chronic stall gave downstreamAnnounceIWF enough time to push a stale, pre-transition clockClass 6 (LOCKED) to the downstream ptp4l config after the BC had already correctly moved to HOLDOVER and announced clockClass 135.

Fixes the regex to accept hex clock identities, adds a BC-state re-check immediately before each downstream PMC write as a race guard, and adds regression tests against real captured pmc output.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented outdated timing data from being written after the clock state changes.
  * Improved validation of clock identity and parent data received from timing services.
  * Fixed parsing for valid hexadecimal clock identities, including uppercase values.

* **Tests**
  * Added comprehensive coverage for timing data parsing, validation, updates, and edge cases.

### DIFF
--- a/pkg/event/event_tbc.go
+++ b/pkg/event/event_tbc.go
@@ -457,11 +457,22 @@ func (e *EventHandler) downstreamAnnounceIWF(ctx context.Context, cfgName string
 	}
 	glog.Infof("%++v", es)
 	e.announceClockClass(gs.ClockQuality.ClockClass, gs.ClockQuality.ClockAccuracy, cfgName)
-	if err := pmc.SetExternalGMPropertiesNP(controlledPortsConfig, es); err != nil {
-		glog.Error(err)
+
+	// Re-check we're still LOCKED right before each write: gs/es are a stale snapshot if a slow PMC call let the BC move to HOLDOVER/FREERUN in the meantime.
+	if !e.applyIfLockedBC(cfgName, "before downstream PMC writes", func() {
+		if err := pmc.SetExternalGMPropertiesNP(controlledPortsConfig, es); err != nil {
+			glog.Error(err)
+		}
+	}) {
+		return
 	}
-	if err := pmc.SetGMSettings(controlledPortsConfig, gs); err != nil {
-		glog.Error(err)
+
+	if !e.applyIfLockedBC(cfgName, "before SetGMSettings", func() {
+		if err := pmc.SetGMSettings(controlledPortsConfig, gs); err != nil {
+			glog.Error(err)
+		}
+	}) {
+		return
 	}
 	glog.Infof("%++v", es)
 
--- a/pkg/protocol/type.go
+++ b/pkg/protocol/type.go
@@ -18,6 +18,17 @@ const (
 	ClockClassOutOfSpec     protocol.ClockClass = 140
 )
 
+// PMC field-name keys for ExternalGrandmasterProperties, shared across
+// ValueRegEx/Keys/Update (and reused by tests) to avoid duplicate literals.
+const (
+	keyGMIdentity   = "gmIdentity"
+	keyStepsRemoved = "stepsRemoved"
+)
+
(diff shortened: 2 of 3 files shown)


pkg/event/event_tbc.go
```diff
@@ -407,6 +407,9 @@ func (e *EventHandler) applyIfLockedBC(cfgName, context string, fn func()) bool
 	fn()
 	return true
 }
+func (e *EventHandler) isLockedBC(cfgName, context string) bool {
```
nocturnalastro: Refactor out an isLocked() then make applyIfLocked Use it

pkg/event/event_tbc.go
```diff
@@ -457,9 +460,18 @@ func (e *EventHandler) downstreamAnnounceIWF(ctx context.Context, cfgName string
 	}
 	glog.Infof("%++v", es)
 	e.announceClockClass(gs.ClockQuality.ClockClass, gs.ClockQuality.ClockAccuracy, cfgName)
+
+	// Re-check we're still LOCKED right before each write: gs/es are a stale snapshot if a slow PMC call let the BC move to HOLDOVER/FREERUN in the meantime.
+	if !e.isLockedBC(cfgName, "before downstream PMC writes") {
+		return
+	}
 	if err := pmc.SetExternalGMPropertiesNP(controlledPortsConfig, es); err != nil {
```
nocturnalastro: Or use the same pattern as vitaly. 
```go
if !e.applyIfLockedBC(cfgName, "before downstream PMC writes", func(){
      if err = pmc.SetExternalGMPropertiesNP(controlledPortsConfig, es); err != nil {
          glog.Error(err)
      }
}) {
		return
	}
```

pkg/protocol/type.go
```diff
@@ -18,6 +18,21 @@ const (
...
+	keyStepsRemoved = "stepsRemoved"
+)
+
+// clockIdentityPattern matches a PTP clockIdentity as printed by ptp4l/pmc:
+// an 8-octet EUI-64 value formatted as three hex groups (e.g.
+// 507c6f.fffe.1fb16c). Shared by every DataSet field that carries a bare
+// clock identity (ExternalGrandmasterProperties.gmIdentity,
+// ParentDataSet.grandmasterIdentity) so they all stay consistent and
+// correctly reject non-hex/malformed values instead of matching anything.
+const clockIdentityPattern = `[0-9a-fA-F]+\.[0-9a-fA-F]+\.[0-9a-fA-F]+`
```
nocturnalastro, line 30: Would be better if the GM identity stuff was a seperate commit. Then you need less stuff in comments and more in git log.

This should just be more like
```
// clock identities are an 8-octet EUI-64 value formatted as three hex groups (e.g.
// 507c6f.fffe.1fb16c). 
```
explaining why it needs to be its own pattern is superfluous

pkg/event/event_tbc.go
gtannous-spec: Reverted as it was before fixing the nitpick comment !

pkg/event/event_tbc.go
gtannous-spec: Fixed.

pkg/protocol/type.go
gtannous-spec: Done ✅

gtannous-spec:
@nocturnalastro Addressed the comments :) 
let me know if it looks better and/or need any other changes from my side !

lack (approved):

PR #246, opened 2026-07-13
T-GM support for v2 API and HardwareConfig

lack:
- **Add GNSS source initialization and T-GM behavior templates**
- **Skip storing empty command slices in resolveClockChainBehavior**
- **Skip upstream port derivation for T-GM clock type**
- **Add intel/e810 T-GM behavior profile template**

## Summary by CodeRabbit

* **New Features**
  * Added T-GM clock profile support with updated vendor behavior templates (GNSS time source + “Initialize T-GM” trigger).
  * Added HardwareConfig-driven GNSS discovery and automatic GPSD GNSS init command generation (antenna voltage, batched constellation enable/disable, optional survey-in).
* **Bug Fixes**
  * Improved clock-chain resolution and template/user behavior merging to always derive template behavior and correctly compute interface/DPLL details.
  * When GNSS serial can’t be resolved, continues using the default GPSD GNSS serial port.
* **Tests**
  * Expanded unit tests for T-GM resolution, behavior merging, and GNSS init command ordering.
* **Configuration**
  * Refreshed Intel E810 defaults and updated pin behavior in vendor templates.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -25,6 +25,7 @@ import (
 	ptpnetwork "github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/network"
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/parser"
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/synce"
+	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/ublox"
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/utils"
 	ptpv2alpha1 "github.com/k8snetworkplumbingwg/ptp-operator/api/v2alpha1"
 	ptpclient "github.com/k8snetworkplumbingwg/ptp-operator/pkg/client/clientset/versioned"
@@ -1293,6 +1294,21 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) 
 			glog.Infof("Delaying phc2sys startup: %s", dprocess.skipInitialStartup)
 		} else if pProcess == ts2phcProcessName { //& if the x plugin is enabled
 			if clockType == event.GM {
+				// If a HardwareConfig defines a GNSS source, locate the serial port and any GNSS initialization commands.
+				var gnssInitCmds ublox.CommandList
+				if nodeProfile.Name != nil && dn.hardwareConfigManager.ReadyHardwareConfigForProfile(*nodeProfile.Name) {
+					gnssPort, gnssErr := dn.hardwareConfigManager.GetGNSSSerialPort(nodeProfile)
+					if gnssErr != nil {
+						glog.Warningf("HardwareConfig GNSS device detection failed: %v", gnssErr)
+					} else if gnssPort != "" {
+						glog.Infof("HardwareConfig GNSS device detected: %s (overriding ts2phc config)", gnssPort)
+						output.gnss_serial_port = gnssPort
+					}
+
+					gnssInitCmds = dn.hardwareConfigManager.GetGNSSInitCommands(nodeProfile)
+					glog.Infof("HardwareConfig GNSS initialization added %d additional commands", len(gnssInitCmds))
+				}
+
 				if output.gnss_serial_port == "" {
 					output.gnss_serial_port = GPSDDefaultGNSSSerialPort
 					glog.Warningf("Setting GNSS serial port to %s", output.gnss_serial_port)
@@ -1302,6 +1318,7 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) 
 
 				// Record ublox init results (MON-HW, etc.) to NodePtpDevice status.
 				// Replaces any previous "gnss" entries to avoid duplicates on re-init.
+				// Additionally store on HardwareConfigManager when that path is active.
 				gnssResultsFn := func(results []string) {
 					dn.hwconfigsMu.Lock()
 					defer dn.hwconfigsMu.Unlock()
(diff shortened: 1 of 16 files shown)


pkg/hardwareconfig/clockchain_resolution.go
```diff
@@ -495,17 +532,17 @@ func deepCopyAndResolveCondition(cond ptpv2alpha1.Condition, vars templateVariab
...
+	if clockChain.Behavior != nil {
+		userSources = clockChain.Behavior.Sources
 	}
 
-	// Initialize behavior if nil
-	if clockChain.Behavior == nil {
-		clockChain.Behavior = &ptpv2alpha1.Behavior{}
-	}
+	// Initialize/reset behavior for template derivation
+	clockChain.Behavior = &ptpv2alpha1.Behavior{}
```
vitus133, line 547: We save sources, but drop Conditions when resetting behavior here

pkg/hardwareconfig/hardwareconfig.go
```diff
@@ -428,18 +428,17 @@ func (hcm *HardwareConfigManager) HasHardwareConfigForProfile(nodeProfile *ptpv1
...
-	for _, hwConfig := range hcm.hardwareConfigs {
+	var profiles []*ptpv2alpha1.HardwareProfile
+	for i, hwConfig := range hcm.hardwareConfigs {
 		if ProfileNamesMatch(*nodeProfile.Name, hwConfig.Spec.RelatedPtpProfileName) {
-			relevantConfigs = append(relevantConfigs, hwConfig.Spec.Profile)
+			profiles = append(profiles, &hcm.hardwareConfigs[i].Spec.Profile)
 		}
 	}
-	return relevantConfigs
+	return profiles
```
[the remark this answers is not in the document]
vitus133: Changing return type to pointers makes the profiles mutable by callers. Returning by value felt safer. Any specific reason for this change?

pkg/daemon/daemon.go
```diff
@@ -1293,15 +1294,36 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
...
 				if output.gnss_serial_port == "" {
 					output.gnss_serial_port = GPSDDefaultGNSSSerialPort
 					glog.Warningf("Setting GNSS serial port to %s", output.gnss_serial_port)
 				}
 				gmInterface := dprocess.ifaces.GetLeadingInterface().Name
 				glog.Infof("Working with GNSS serial port %s, leading iface %s", output.gnss_serial_port, gmInterface)
 
+				// Get GNSS init commands from HardwareConfig if available
+				var gnssInitCmds ublox.CommandList
+				hasHwConfig := nodeProfile.Name != nil && dn.hardwareConfigManager.ReadyHardwareConfigForProfile(*nodeProfile.Name)
```
vitus133: Also called identically on line 1299 - can save this call

pkg/hardwareconfig/clockchain_resolution.go
```diff
@@ -93,30 +93,53 @@ func extractUpstreamPortsFromPtpProfile(ptpProfile *ptpv1.PtpProfile) []string {
 	return upstreamPorts
 }
 
-// findLeadingInterfaceFromUpstreamPort finds the leading interface (for DPLL clock ID) from an upstream port.
+// extractInterfacesFromPtpProfile extracts all interface names from a PTP profile's
+// ptp4l config (any [section] that is not a well-known non-interface section).
+func extractInterfacesFromPtpProfile(ptpProfile *ptpv1.PtpProfile) []string {
```
vitus133, line 98: nit: We have something similar in `PopulatePtp4lConf`, and we can't reuse it  because of circular dependency. I think we have a tech debt story around this. Just complaining for visibility :)

pkg/hardwareconfig/hardwareconfig.go
```diff
@@ -631,14 +630,18 @@ func (hcm *HardwareConfigManager) resolveClockChainBehavior(hwConfig ptpv2alpha1
 			return nil, nil, fmt.Errorf("failed to resolve DPLL pin commands for condition %s: %w", condition.Name, err)
 		}
 		glog.Infof("    Condition '%s': resolved %d DPLL commands", condition.Name, len(pinCommands))
-		pinCommandsPerCondition[condition.Name] = pinCommands
+		if len(pinCommands) > 0 {
```
vitus133, line 633: This wouldn't store empty maps now. It means that when there is no behavior defined (as in T-GM), nothing will be cached. Every time the condition occurs, a warning will be printed in ApplyConditionForProfile that there are no cached commands

vitus133 (commented):
In general looks good! I commented on a couple of concerns

pkg/daemon/daemon.go
lack: They were added separately at first, so each one stands (stood) alone.  But yes, we could simplify that.

pkg/hardwareconfig/clockchain_resolution.go
lack: Yes, having a general programatic wrapper around ptp4l.conf and ts2phc.conf would be a nice thing to engineer.  We have a lot of bespoke parsing here an there.

pkg/hardwareconfig/hardwareconfig.go
lack: I added these guards because I was seeing confusing logging.  The caller says there was 1 sysfs condition when I knew there were none.  Reason was because there was one entry with an empty list...

lack: But I'll take a closer look and see if it's still needed.  I've gone through a few refactors since then.

pkg/hardwareconfig/clockchain_resolution.go
lack: Indeed.  I'll repair that.

pkg/hardwareconfig/hardwareconfig.go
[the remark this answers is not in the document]
lack: I think Claude slid this in because he likes the efficiency of pointers, but I agree it's better to keep pass-by-value.

lack:
Manual testing completed:
- T-GM on WPC with HardwareConfig v2 (initializes and locks)
- T-GM on GNR-D with HardwareConfig v2 (initializes and locks)
- T-BC on GNR-D with HardwareConfig v2 (initializes, locks, holdover, re-locks)

vitus133 (approved):

PR #251, opened 2026-07-21
Cleanup GNR-D Hwconfig documentation and defaults

lack:
- **Add t-gm for GNR-D documentation**
- **Add reasonable GNSS TTY device defaults for hardwareconfigv2 behavior profiles**

## Summary by CodeRabbit

- **New Features**
  - Added documentation with end-to-end configuration examples for T-GM GNSS holdover on Dell XR8720t, including sample profiles and Kubernetes/OpenShift resources.
- **Bug Fixes**
  - Ensured GNSS device matcher settings are preserved during configuration merging, even when the user omits the device match.
- **Tests**
  - Updated GNSS tests to use a shared `/dev/ttyACM0` constant and expanded coverage with table-driven scenarios.

## Testing Done

Ensured that T-GM goes locked (state s2) on GNR-D with the example config in `doc/t-gm-dell-XR8720t.md`

### DIFF
--- a/pkg/hardwareconfig/clockchain_resolution.go
+++ b/pkg/hardwareconfig/clockchain_resolution.go
@@ -660,6 +660,17 @@ func mergeSourceConfig(tpl, user *ptpv2alpha1.SourceConfig) {
 		tpl.PTPTimeReceivers = user.PTPTimeReceivers
 	}
 	if user.GNSSConfig != nil {
-		tpl.GNSSConfig = user.GNSSConfig
+		// Explicitly shallow-copy the user GNSS config
+		gnss := *user.GNSSConfig
+		if user.GNSSConfig.Match != nil {
+			// Prefer: shallow-copy the user GNSS matcher (if set)
+			match := *user.GNSSConfig.Match
+			gnss.Match = &match
+		} else if tpl.GNSSConfig != nil && tpl.GNSSConfig.Match != nil {
+			// Fallback: shallow-copy the original template matcher (if set)
+			match := *tpl.GNSSConfig.Match
+			gnss.Match = &match
+		}
+		tpl.GNSSConfig = &gnss
 	}
 }
--- a/doc/t-gm-dell-XR8720t.md
+++ b/doc/t-gm-dell-XR8720t.md
@@ -43,7 +43,7 @@ spec:
         verbose 1
         logging_level 7
         ts2phc.pulsewidth 100000000
-        ts2phc.nmea_serialport /dev/ttyACM0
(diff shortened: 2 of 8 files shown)


pkg/hardwareconfig/clockchain_resolution.go
```diff
@@ -660,6 +660,10 @@ func mergeSourceConfig(tpl, user *ptpv2alpha1.SourceConfig) {
 		tpl.PTPTimeReceivers = user.PTPTimeReceivers
 	}
 	if user.GNSSConfig != nil {
+		if user.GNSSConfig.Match == nil && tpl.GNSSConfig != nil {
+			// Default to the template match section if not specified by the user (may be nil)
+			user.GNSSConfig.Match = tpl.GNSSConfig.Match
+		}
 		tpl.GNSSConfig = user.GNSSConfig
```
sebsoto: This is going to mutate the passed in user config, which makes it an outlier from the rest of the function which only mutates tpl.

You could preserve the previous behavior of only modifying the passed in template like this:
```
  if user.GNSSConfig != nil {
      gnss := *user.GNSSConfig
      if gnss.Match == nil && tpl.GNSSConfig != nil {
          match := *tpl.GNSSConfig.Match
          gnss.Match = &match
      } else if gnss.Match != nil {
          match := *gnss.Match
          gnss.Match = &match
      }
      tpl.GNSSConfig = &gnss
  }
```

Alternatively, a copy of the template could be created and merge the user fields onto it, which would mutate neither. This could make things a bit cleaner, but might be out of scope of this change

sebsoto (changes requested):
Just one comment otherwise LGTM!

pkg/hardwareconfig/clockchain_resolution.go
```diff
@@ -660,6 +660,17 @@ func mergeSourceConfig(tpl, user *ptpv2alpha1.SourceConfig) {
...
 	}
 	if user.GNSSConfig != nil {
-		tpl.GNSSConfig = user.GNSSConfig
+		// Explicitly shallow-copy the user GNSS config
+		gnss := *user.GNSSConfig
+		if gnss.Match != nil {
+			// Prefer: shallow-copy the user GNSS matcher (if set)
+			match := *gnss.Match
+			gnss.Match = &match
+		} else if tpl.GNSSConfig != nil {
```
sebsoto: We need to do the nil check on match too right?
`else if tpl.GNSSConfig != nil && tpl.GNSSConfig.Match != nil`?

Missed this on my original suggestion

sebsoto (approved):

PR #254, opened 2026-07-23
CNF-26080: Add ipc cache for cloud event proxy v2

sebsoto:
This is not hooked up to the daemon yet, and thus will not impact existing functionality.

Adds a cache to store the latest state of all clocks watched by the daemon. As new state events come into the cache they will be pushed to the cloud-event-proxy v2 socket.

Allows a new instance of cloud event proxy v2 to get the current state of the system by querying the daemon.

## Summary by CodeRabbit

* **New Features**
  * Added automatic daemon state synchronization on IPC connect via status requests and on-demand snapshots.
  * Introduced IPC message caching with duplicate suppression and buffered forwarding.
  * Improved IPC socket communication with reconnect behavior and live forwarding.
  * Extended state payload to support an optional offset field.
* **Bug Fixes**
  * Status updates are now retained and forwarded instead of being discarded.
* **Tests**
  * Updated unit, integration, and end-to-end tests, including signal-driven IPC sender flows and snapshot verification via polling.

### DIFF
--- a/pkg/cep/cep_test.go
+++ b/pkg/cep/cep_test.go
@@ -134,10 +134,10 @@ func TestListenerResilience(t *testing.T) {
 			name: "version mismatch ignored",
 			input: func() io.Reader {
 				var buf bytes.Buffer
-				ipc.Encode(&buf, []ipc.Message{{
+				ipc.Transmit(&buf, ipc.Message{
 					Version: 99, Type: ipc.TypePTPState, Profile: testProfile,
 					IFace: testIFace, Values: ipc.StateValue{State: ipc.StateLocked},
-				}})
+				})
 				return &buf
 			},
 			wantCached: false,
@@ -147,7 +147,7 @@ func TestListenerResilience(t *testing.T) {
 			input: func() io.Reader {
 				var buf bytes.Buffer
 				buf.WriteString("not json\n")
-				ipc.Encode(&buf, []ipc.Message{validMsg})
+				ipc.Transmit(&buf, validMsg)
 				return &buf
 			},
 			wantCached: true,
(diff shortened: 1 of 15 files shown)


pkg/cep/cep_test.go
```diff
@@ -463,12 +449,12 @@ func TestIntegrationCurrentState(t *testing.T) {
...
-	ipc.Encode(pw, []ipc.Message{
-		{Version: ipc.Version, Type: ipc.TypePTPState, Profile: testProfile,
-			IFace: testIFace, Values: ipc.StateValue{State: ipc.StateLocked}},
-		{Version: ipc.Version, Type: ipc.TypePTPState, Profile: testProfile,
-			IFace: testIFace, Values: ipc.StateValue{State: ipc.StateFreerun}},
-	})
+	ipc.Transmit(pw, ipc.Message{
+		Version: ipc.Version, Type: ipc.TypePTPState, Profile: testProfile,
+		IFace: testIFace, Values: ipc.StateValue{State: ipc.StateLocked}})
+	ipc.Transmit(pw, ipc.Message{
```
nocturnalastro, line 455: Is there a reason Trasmit isn't variadic? 

That way it would just be `ipc.Transmit(pw, msg1, msg2)`

pkg/ipc/cache.go
```diff
@@ -0,0 +1,94 @@
...
+	defer c.mu.Unlock()
+	c.store = make(map[cacheKey]Message)
+}
+
+// Out returns the read-only outbound message channel.
+func (c *Cache) Out() <-chan Message {
+	return c.outCh
+}
+
+func valuesEqual(a, b Message) bool {
```
nocturnalastro: Feels like this should be a function on message but I'm not that bothered either way.

pkg/ipc/cache.go
```diff
@@ -0,0 +1,94 @@
...
+	if a.IFace != b.IFace {
+		return false
+	}
+	switch a.Values.(type) {
```
nocturnalastro: Do we need to check if both a and b are the same value type?

pkg/ipc/link.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+}
+
+// dial retries indefinitely with exponential backoff until a connection
+// is established or the context is canceled.
+func (l *Link) dial(ctx context.Context) net.Conn {
+	dialFn := func() (net.Conn, error) { return l.dialFn(ctx) }
+
+	for {
+		conn := utils.ReconnectWithBackoff(ctx, dialFn, utils.DefaultReconnectConfig())
```
nocturnalastro: How do we get the error if this is failing?

pkg/ipc/link.go
```diff
@@ -0,0 +1,152 @@
...
+		default:
+			glog.Warning("IPC link: connection to socket timed out, retrying")
+		}
+	}
+}
+
+// scan reads incoming messages from conn and sends them on ch.
+// It closes ch when the scanner fails, signaling that the connection
+// is broken.
+func (l *Link) scan(ctx context.Context, conn net.Conn, ch chan<- Message) {
```
nocturnalastro: could this just be a reader instead of net.Conn? might make mocking easier.

pkg/ipc/link.go
```diff
@@ -0,0 +1,152 @@
...
+		select {
+		case ch <- msg:
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+// serve writes data to the socket as required
+func (l *Link) serve(ctx context.Context, conn net.Conn, statusCh <-chan Message) {
```
nocturnalastro, line 106: Same here can it just be a writer?

pkg/ipc/link.go
sebsoto: Good point, changed

pkg/ipc/cache.go
sebsoto: Yeah that makes more sense

pkg/cep/cep_test.go
sebsoto: Added for convenience :+1:

pkg/ipc/link.go
sebsoto: This needs to be a conn for the SetDeadlline call

nocturnalastro (approved):
/lgtm

PR #257, opened 2026-07-28
CNF-19182: Add PTP sync status reporting to NodePtpDevice CR (Phase 1)

aabughosh:
## Summary
Implements Phase 1 of [CNF-19182](https://issues.redhat.com/browse/CNF-19182): lightweight PTP status reporting via NodePtpDevice Custom Resource conditions and a synchronization snapshot.
This focuses on non-metrics status that Kubernetes controllers and GitOps tooling need — fields that duplicate Prometheus metrics (clock state, clock class, process liveness) are intentionally omitted per reviewer feedback on [ptp-operator#277](https://github.com/k8snetworkplumbingwg/ptp-operator/pull/277#pullrequestreview-4785661475).
## Changes
1. **vendor: add SyncStatus and Conditions to NodePtpDevice status**
   - Add `SyncStatus` struct with `ClockTypes []string`, `AppliedProfile`, and `Interfaces`
   - Add `Sync *SyncStatus` and `Conditions []metav1.Condition` to `NodePtpDeviceStatus`
2. **event: add sync status transition notification channel**
   - Add `syncStatusCh` to `EventHandler`, signalled (non-blocking) on GM and T-BC/T-TSC clock state transitions
   - Avoids etcd write amplification by only triggering updates on meaningful state changes
3. **daemon: add runSyncStatusUpdate for NodePtpDevice CR**
   - Collects `clockTypes` (all unique clock types from active processes as plain strings — supports multi-profile nodes)
   - Populates `appliedProfile` and sorted `interfaces` list
   - Sets `Ready` condition: True when a profile is applied and processes are running
   - Uses conflict retry (3 attempts) on `UpdateStatus` to handle races with `runDeviceStatusUpdate()`
   - Backfills `devices[].profile` from active process list
4. **daemon: wire sync status goroutine with heartbeat**
   - Triggers `runSyncStatusUpdate()` on clock state transitions and every 30s (periodic refresh)
## Design Decisions
- **`clockTypes []string` (not a typed enum):** Supports multiple simultaneous clock roles (e.g. two BC profiles, or GM + BC on the same node) and future-proofs against new clock types without API changes.
- **Only `Ready` condition (no `Synced`):** Clock state is already available via Prometheus metrics; duplicating it in the CR adds write churn without new value for controllers.
- **No `lastTransitionTime`/`lastUpdateTime` on SyncStatus:** Removed per reviewer feedback — the standard `Condition.LastTransitionTime` is sufficient.
## Depends On
- [ptp-operator#277](https://github.com/k8snetworkplumbingwg/ptp-operator/pull/277) — upstream API type changes (vendored locally until merged)
## Test Plan
- [ ] Build passes (`make`)
- [ ] Unit tests pass (`go test ./pkg/daemon/... ./pkg/event/...`)
- [ ] Deploy with test image and verify `NodePtpDevice.status.sync` is populated
- [ ] Verify `Ready` condition transitions correctly on profile apply/remove
- [ ] Verify no status updates on every offset tick (only on state transitions + 30s heartbeat)

Jira: https://issues.redhat.com/browse/CNF-19182

## Summary by CodeRabbit

* **New Features**
  * Added automatic synchronization status updates when clock states change.
  * Sync status now refreshes periodically and reflects the applied profile, clock type, active interfaces, and device profiles.
  * Added readiness reporting based on whether a profile is applied and processes are running.

* **Bug Fixes**
  * Improved status update reliability by retrying temporary update conflicts.
  * Preserved readiness transition times when the reported status remains unchanged.


pkg/daemon/daemon.go
```diff
@@ -1548,6 +1571,7 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 
 	}
 	glog.Infof("Completed applyNodePtpProfile for profile %s, total processes in manager: %d", *nodeProfile.Name, len(dn.processManager.process))
+	dn.runSyncStatusUpdate()
```
nocturnalastro: I don't think this one is needed. Just do the one in `applyNodePTPProfiles`

pkg/daemon/daemon.go
```diff
@@ -703,6 +665,28 @@ func (dn *Daemon) Run() {
 	glog.Info("Daemon Run() started, waiting for configuration updates...")
 	go dn.processManager.ptpEventHandler.ProcessEvents()
 
+	// Watch for clock-state transitions and emit a periodic heartbeat so that
+	// NodePtpDevice.status.sync stays current without flooding etcd on every
+	// offset tick. The goroutine exits when dn.stopCh is closed.
+	go func() {
+		syncStatusCh := dn.processManager.ptpEventHandler.SyncStatusUpdateCh()
+		heartbeat := time.NewTicker(30 * time.Second)
```
nocturnalastro: I'm not sure a heartbeat would be that beneficial. Status should be generally stable anyway. We have events for people who need updates.

I'm not totally against it either, I'm not sure if its worth it.

pkg/daemon/ptpdev.go
```diff
@@ -166,6 +176,151 @@ func RunDeviceStatusUpdate(ptpClient *ptpclient.Clientset, nodeName string, hwco
 	runDeviceStatusUpdate(ptpClient, nodeName, hwconfigs)
 }
 
+// runSyncStatusUpdate writes non-metrics PTP status into NodePtpDevice.status.sync
+// and the Ready condition. Triggered on state transitions and periodic heartbeat.
+func (dn *Daemon) runSyncStatusUpdate() {
+	if dn.ptpClient == nil {
```
nocturnalastro, line 193: We probably a check on the `applyingProfiles` as updating the summary while we are apply the profiles would just cause more writes and maybe even concurrent ones. So we also probably want form of lock so only one thing can be updating at the same.

pkg/daemon/daemon.go
aabughosh: Agreed  dropped it

pkg/daemon/ptpdev.go
aabughosh: Done

pkg/daemon/daemon.go
```diff
@@ -499,6 +461,12 @@ type Daemon struct {
 	ptpClient      *ptpclient.Clientset
 	unknownPlugins []string
 
+	// syncStatusMu serializes NodePtpDevice status.sync writes so apply and
+	// event-driven updates cannot overlap. syncStatusWG tracks in-flight
+	// writers so apply can wait for them before tearing down processes.
+	syncStatusMu sync.Mutex
+	syncStatusWG sync.WaitGroup
```
nocturnalastro: I'm not sure the wait group is needed if we just want a single running update. We can just use the lock itself.

pkg/daemon/ptpdev_internal_test.go
```diff
@@ -0,0 +1,509 @@
...
+			proc: &ptpProcess{
+				clockType: event.BC,
+				nodeProfile: ptpv1.PtpProfile{
+					PtpSettings: map[string]string{"clockType": TBC},
+				},
+			},
+			want: TBC,
+		},
+		{
+			name: "ptpSettings clockType ignored if not T-GM or T-BC",
```
nocturnalastro: This isn't really true profile clock type won't be set to anything other than T-GM or T-BC.

Also this test sets both clock types to the same value so we wouldn't be able to tell.

pkg/daemon/daemon.go
aabughosh: dropped it

pkg/daemon/ptpdev_internal_test.go
aabughosh: Removed

pkg/daemon/ptpdev.go
```diff
@@ -166,6 +176,169 @@ func RunDeviceStatusUpdate(ptpClient *ptpclient.Clientset, nodeName string, hwco
...
+		dn.processManager.ptpEventHandler != nil &&
+		dn.processManager.ptpEventHandler.IsApplying()
+}
+
+func (dn *Daemon) doSyncStatusUpdate(force bool) {
+	if dn.ptpClient == nil {
+		return
+	}
+
+	dn.syncStatusMu.Lock()
```
nocturnalastro, line 197: Would it be worth using a TryLock so that if one is already running we can just exit or should we re-run the update in case something change in status that has already been processed?

nocturnalastro: NVM I think its better to update more for now.

nocturnalastro (dismissed):
/lgtm

pkg/clockmgr/clockmgr.go
```diff
@@ -232,6 +253,10 @@ func (m *ClockManager) ProcessEvents(ctx context.Context) {
 				}
 			}
 			clockState := clk.AddEvent(ev)
+			if prev, hasPrev := m.clockStates[lookupName]; hasPrev && prev != clockState.State {
```
nocturnalastro: I think we need a function to look this up on the clock itself so we can do something like 
```go
dn.processManager.clockMgr.GetClock(name).GetState()
```
or have whatever signals the channel pass you the value for the updated state?

nocturnalastro (approved):
/lgtm

PR #260, opened 2026-07-28
Add Dockerfile for CEPv2

sebsoto:
## Summary by CodeRabbit

* **New Features**
  * Added containerized build and runtime support for the Cloud Event Proxy.
  * Added automatic generation of the Cloud Event Proxy binary as part of the standard build.
  * Embedded build commit metadata and now logs it during startup.

* **Build/Packaging**
  * Updated the container image to use a multi-stage build with a non-root runtime user and proper runtime directory setup.

* **Tests**
  * Revised the Cloud Event Proxy end-to-end test setup to use the standard build output for the proxy binary.


nocturnalastro (approved):
/lgtm

PR #262, opened 2026-08-05
OCPBUGS-7811: preserve process_restart_count across profile teardown

bnshr:
## Summary
- Stop deleting `openshift_ptp_process_restart_count` during process/config teardown in `deleteProcessStatusMetrics` / `DeleteProcessStatusMetrics`.
- Still delete `openshift_ptp_process_status` so scrapers do not see a stale UP/DOWN sample after teardown (OCPBUGS-7808).
- Add a unit test that models teardown + re-apply and asserts the counter stays monotonic (OCPBUGS-7811).

`applyNodePTPProfiles` always tears processes down before re-applying the same config slot (e.g. `ts2phc.0.config`). Deleting the restart counter removed the Prometheus series, so the next `Inc()` started from zero — observed in ptp-operator T-GM conformance as `restart_count` dropping (e.g. 9→5) after GM `PtpConfig` delete/re-apply.

## Test plan
- [x] `go test ./pkg/daemon/ -run TestProcessRestartCount -count=1`
- [ ] Confirm ptp-operator serial test "Verify ts2phc metrics are cleaned up after GM PtpConfig delete and re-apply" passes against a daemon image built from this branch

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved cumulative process restart counts when process status metrics are removed and later restored.
  * Restart counters now continue from their previous value instead of resetting after process or configuration teardown.

* **Tests**
  * Added coverage validating restart-count preservation and continued incrementing across teardown and reapplication.

### DIFF
--- a/pkg/daemon/metrics.go
+++ b/pkg/daemon/metrics.go
@@ -649,11 +649,16 @@ func deleteOsClockStateMetrics(profiles map[string][]string) {
 }
 
 func deleteProcessStatusMetrics(config, process string) {
+	// Remove the process_status gauge when the process/config is torn down so
+	// scrapers do not see a stale UP/DOWN sample (OCPBUGS-7808).
+	//
+	// Do NOT delete process_restart_count: it is a cumulative counter keyed by
+	// (process, node, config). applyNodePTPProfiles always stopAllProcesses
+	// before re-applying the same config slot (e.g. ts2phc.0.config); deleting
+	// the series resets the counter and breaks monotonicity across delete/re-apply
+	// (OCPBUGS-7811). Process crash/restart already continues the counter via Inc().
 	ProcessStatus.Delete(prometheus.Labels{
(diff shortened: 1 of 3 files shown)


bnshr:
Addressed CodeRabbit findings that still applied:

**Fixed**
- **Exercise production cleanup path** (justified): test now calls `deleteProcessStatusMetrics` / `UpdateProcessStatusMetrics` on the real `ProcessStatus` / `ProcessRestartCount` vectors. Verified that re-adding `ProcessRestartCount.Delete` fails with `got 0 want 9`.
- **Assert legacy reset exactly 5** (minor, valid): tightened the documentation test.

**Already resolved / not applicable**
- **goconst `"config"` / err shadowing / `Gather` error**: fixed earlier by switching to `testutil` (`bcdd0ce`).
- **Docstring coverage warning**: skipped — production cleanup already has comments; this is a test-only change, not missing API docs.

